PR #3, opened 2026-07-24
feat: support Hysteria2 port hopping

TnZzZHlp:
## Summary

- Parse Hysteria2 multi-port share-link authorities such as `hy2://…:443,8443,40000-50000`.
- Keep the QUIC connection alive while switching to a marked UDP socket on a new HY2 port every 30 seconds.
- Add parser and transport coverage, plus English and Chinese documentation.

## Why

`url::Url` rejects Hysteria2's multi-port authority syntax, and a QUIC client needs peer-port normalization while migrating sockets. Without both pieces, HY2 port-hopping links cannot be used reliably.

## Validation

- `cargo test -q -p honk-config share_link --lib`
- `cargo test -q -p honk-outbound hysteria2 --lib`
- `cargo test -q -p honk-outbound port_hopping --lib`
- `cargo check -q -p honk-core --features ebpf`
- End-to-end HY2 test: `curl https://ip.sb` succeeded before and after two live port migrations.

### DIFF
--- a/crates/honk-outbound/src/quic.rs
+++ b/crates/honk-outbound/src/quic.rs
@@ -17,22 +17,24 @@
 //!   [`SendStream`]/[`RecvStream`] so it can be boxed into a
 //!   [`crate::proxy::ProxyStream`].
 
-use std::io;
+use std::io::{self, IoSliceMut};
 use std::net::SocketAddr;
 use std::pin::Pin;
 use std::sync::Arc;
+use std::sync::atomic::{AtomicU16, Ordering};
 use std::task::{Context, Poll};
 use std::time::{Duration, Instant};
 
 use anyhow::{Context as _, anyhow};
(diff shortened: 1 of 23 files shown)


crates/honk-outbound/src/quic.rs
```diff
@@ -303,10 +564,144 @@ impl<C> QuicClient<C> {
...
+                Err(e) => {
+                    hopper.restore(old_index);
+                    warn!(
+                        old_port,
+                        target_port, "Hysteria2 port hop skipped: failed to create UDP socket: {e}"
+                    );
+                    continue;
+                }
+            };
+            if let Err(e) = endpoint.rebind_abstract(socket) {
```
Glassyiris: The per-hop rebind_abstract with a fresh socket feels a bit heavy for something that happens every 30s.

PR #40, opened 2026-08-13
fix(urltest): align session protocol latency timing

Glassyiris:
**using warmable node to do urltest**

## What this does

Dashboard delay numbers were misleading for session-based protocols — a cold AnyTLS or
Hysteria2 node would report its full handshake + session setup as "latency", which isn't
what your traffic actually experiences once the connection is up. This PR makes delay
tests measure what sing-box measures: the warm-path RTT.

## How it works

- Node already warm → measure through the existing runtime, same as before. Nothing changes.
- Node cold → spin up a **throwaway** runtime, warm the session/QUIC client there
  (untimed), measure once, then close it. Group scans still retain zero new state —
  no 500-node scan leaving 500 QUIC clients lying around.
- Hysteria2 gets one extra tweak: it coalesces the target request with the first
  application write (lazy handshake), so timing starts after dial returns, matching
  sing-box's rule. TUIC and Juicity complete the target handshake inside the warmed
  dial instead — all three QUIC protocols now exclude cold connect+auth from the number.

## Heads-up (behavior changes)

- **Displayed delays will drop after upgrading** for AnyTLS / VLESS mux / QUIC nodes.
  The numbers didn't get faster — the measurement got honest.
- Worst case for a cold node is now warm + measure (two round trips of work instead of
  one), still bounded by the same timeout.
- A cold node whose warm-up fails now reports failure without attempting a measurement.

Full semantics documented in `doc/*/reference/api.md` ("Delay measurement").

## Testing

Seven new tests: all five reusable transports warm before measuring, QUIC warm time
stays out of the reported number (100ms fake warm → &lt;50ms reported), warm failure
skips measurement, and the warm-up runtime is verified to be ephemeral.


Glassyiris:
## it will take a new issue

in this time from dashboard | core urltest using different probe system...................

PR #50, opened 2026-08-17
feat(dns): add automatic client subnet presets

Glassyiris:
## Summary

- add `dns.client_subnet` with fixed IPv4/CIDR and `auto[(probe)]` modes
- infer a conservative `/24` from the first public IPv4 route hop and refresh it on reload/network changes
- inject ECS only for eligible upstream queries, preserve caller EDNS, remove synthetic response state, and partition DNS policy/cache identity by the effective subnet
- document the setting in both English and Chinese

## Validation

- `just dns-ci` — all green
- `just build-musl` — x86_64 musl binary with embedded eBPF object
- `10.10.10.49` smoke: automatic inference succeeded; live capture saw the inferred `/24` ECS on upstream requests; listener responses removed the synthetic OPT; external Google DNS resolution succeeded
- VyOS canary `10.10.10.1`: deployed SHA-256 `d4add5e9f63a5ed571e35f4b6c7fff66d8c7de595878beb2a5d3227800eafe0d`, enabled `client_subnet: auto`, restarted `dae.service`, and verified the running executable hash
- real eBPF runtime inferred ECS after interface publication, remained active, served `www.bilibili.com`, `upos-sz-mirrorcos.bilivideo.com`, and `example.com` through `10.10.10.1:5353`, and logged no error-priority entries after deployment

## Canary rollback

- binary: `/config/vyos-scripts/podman/dae/dae.bak-pre-ecs-20260817T094947Z`
- config: `/config/vyos-scripts/podman/dae/config.dae.bak-pre-ecs-20260817T094947Z`


Glassyiris:
Regression follow-up pushed in `33eda30`.

- Reproduced a fixed-flow ECMP failure: one ICMP-silent early path let the original probe select a later `202.97.93.0/24` backbone hop, producing poor CDN locality.
- Automatic inference now sends three independent UDP flows per TTL, still bounded to 12 TTLs and two seconds. Five consecutive gateway restarts selected only first-public access/metro hops (`61.146.244.0/24`, `202.105.154.0/24`, or `202.105.157.0/24`).
- The reload test now proves the replacement DNS generation injects the resolved ECS bytes into a real upstream query, not only that config/policy state changed.
- Validation: focused ECS tests (5 passed), reload wire test (1 passed), `cargo clippy -p honk-config -p honk-core --lib --all-features -- -D warnings`, clean `just build-musl`, and live VyOS rollout.

Live regression result after flushing both DNS caches: Taobao `183.60.205.74` averaged 5.289 ms (bad fallback was 410.699 ms), NetEase 4.907 ms, ByteDance 4.181 ms, and Bilibili 19.905 ms. The local resolver's Cloudflare fallback also had to be replaced with an ECS-capable Google fallback; that was deployment configuration, not a honk code change.

Glassyiris:
Routing semantic follow-up corrected in `094bdf8`.

- Generated ECS is now decided only after the complete upstream dial route resolves, including both explicit `-> tag` selection and ordinary traffic routing.
- A direct attempt, including `-> direct`, receives generated ECS; an attempt resolving to a proxy leaf receives the original query. Thus an unbound `remote: 'tcp+udp://dns.google:53'` routed through a proxy no longer carries the home prefix.
- UDP candidates are judged per attempt. If a direct address fails with ECS and an alternate address routes through a proxy, the retry is rebuilt from the original query without ECS.
- Caller-supplied ECS remains untouched. English and Chinese references now describe the route-coupled behavior.
- Validation: direct-route ECS test (1 passed), mixed direct-to-proxy UDP retry test (1 passed), DNS upstream-pool tests (23 passed), ECS tests (5 passed), client-subnet config tests (5 passed), and full `just dns-ci` passed including fmt, clippy, honk-config, honk-core DNS/control, and honk-outbound suites.

Glassyiris:
Documentation follow-up pushed in `4b70582`.

- The bilingual DNS design now records why proxied DoQ/DoH3 is unavailable: the current proxy dial path yields a TCP byte stream, while quinn requires a datagram socket.
- It records the required implementation boundary: adapt outbound `PacketTransport` to quinn `AsyncUdpSocket` while preserving datagram, MTU, generation, reconnect, and shutdown semantics.
- Both README TODO lists now track proxied DoQ/DoH3 explicitly. Current fail-closed behavior and the DoT/DoH workaround remain documented.

Glassyiris:
Recommended deployment example added in `91676e9`.

`example.dae` now uses the intended split resolver setup:
- AliDNS DoH at a fixed IP, forced through `direct`, with `client_subnet: auto` for domestic CDN locality.
- Google DoH at a fixed IP, forced through the `auto` proxy group; generated ECS is therefore not disclosed on that path.
- `geosite:gfw` goes remote immediately; other queries try AliDNS first, accept private/CN answers, and re-query Google otherwise.
- Both endpoints avoid bootstrap DNS while retaining TLS hostname verification.

Validation: `cargo test -p honk-config --test example_configs` passed. A live UDP ECS probe also confirmed `223.5.5.5` echoes the requested ECS option.

PR #58, opened 2026-08-20
fix(dns): reject truncated upstream responses

Glassyiris:
## Summary

- Centralize DNS TC-bit detection and preserve the full reusable wire for waiters.
- Prevent upstream-truncated responses from cache publication, persistence, answer-IP extraction, or domain-route projection.
- Keep UDP/as-is TCP fallback behavior unchanged.

## Verification

- cargo test -p honk-core --lib --no-default-features --features clash-api
- cargo fmt --all --check
- cargo clippy -p honk-core --lib --no-default-features --features clash-api -- -D warnings


Glassyiris (commented):
check all tests, can approve

PR #67, opened 2026-08-23
fix(routing): derive pname from process argv0

### LINKED ISSUE #64: pname routing uses the current thread name instead of the process name
## Summary

`pname(...)` routing does not reliably match multithreaded applications.

## Why this happens

The current implementation obtains the process name with `bpf_get_current_comm()`.

However, `bpf_get_current_comm()` returns `task_struct->comm`, which is a per-thread name, not the executable name of the process. Cgroup hooks run in the context of the thread performing the socket operation. If an application renames its worker thread, Honk records that worker-thread name in `COOKIE_PID_MAP` while storing the process TGID.

For example:

- Process executable: `python3`
- Process TGID: `<tgid>`
- Worker thread name: `worker-thread`
- Recorded `pname`: `worker-thread`

As a result, a rule such as:

```dae
pname(python3) -> direct(must)
```

does not match sockets created by that worker thread. The traffic falls through to another routing rule or the default route.

## Reproduction

1. Start a process whose executable name is `python3`.
2. Create a worker thread and set its native thread name to `worker-thread`.
3. Create a socket from that worker thread.
4. Configure a `pname(python3)` routing rule.
5. Observe that the rule does not match because the recorded name is `worker-thread`.

This issue is especially visible with multithreaded proxy clients and services that assign names to their worker threads.

## Expected behavior

`pname(...)` should identify the process consistently, regardless of which thread creates the socket. The recorded name should be derived from the process command line, such as the basename of `argv[0]`, rather than from the current thread's `comm`.

This would also match the behavior in dae:

https://github.com/daeuniverse/dae/blob/39831d5d24a129ce8c53258e54cfe6cde1502529/control/kern/tproxy.c#L1760

## Environment

- Linux kernel: 7.1.5
- cgroup v2
- BTF enabled
- eBPF datapath

## Possible fix

Read `task_struct.mm->arg_start` using runtime BTF offsets, read `argv[0]`, and extract the process name from it. `bpf_get_current_comm()` can remain as a fallback when the required helper or BTF information is unavailable.

Glassyiris:
## Summary
- resolve `task_struct.mm` and `mm_struct.arg_start` offsets from runtime kernel BTF
- record the 15-byte `argv[0]` executable basename in `COOKIE_PID_MAP`, with thread `comm` as the safe fallback
- cover renamed worker threads with a real cgroup/eBPF regression and document the routing behavior in both languages

## Verification
- `cargo fmt --all -- --check`
- `env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY cargo test --workspace --no-fail-fast -- --skip test_routing_with_config_dae`
- `env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY cargo test -p honk-core --features ebpf --lib -- --skip test_routing_with_config_dae`
- `just test-netns`
- `cargo clippy -p honk-core --features ebpf --all-targets -- -D warnings`
- `cargo +nightly build --release -Zbuild-std=core --target bpfel-unknown-none` in `crates/honk-ebpf`

Fixes #64


Glassyiris:
## Gateway validation (2026-08-23)

### `10.10.10.49` — positive path

- Arch Linux, x86_64, kernel `7.1.8-arch1-3`, BTF at `/sys/kernel/btf/vmlinux`.
- Cross-built the PR's `honk-core` lib test as a static x86_64-musl executable and ran the exact ignored real-kernel regression:
  - `ebpf::real::tests::pname_uses_process_argv0_from_renamed_thread ... ok`
  - `1 passed; 0 failed`
- This loaded the real cgroup eBPF hooks, created a UDP socket from a thread named `worker-thread`, and verified `COOKIE_PID_MAP.pname` contained the executable basename instead of the thread name.
- The temporary test executable and test pin root were removed.

### `10.10.10.118` — compatibility fallback

- OpenWrt `25.12.5`, aarch64, kernel `6.12.94`; `/sys/kernel/btf/vmlinux` is absent and raw BTF is installed at the newly searched common path `/usr/lib/debug/boot/vmlinux`.
- The file has raw BTF magic `9f eb`; `bpftool` resolves `task_struct.mm` at byte `1048` and nested `mm_struct.arg_start` at byte `352`.
- Deployed the PR's static aarch64-musl binary (`sha256 713b8b835eb9d53098226371a1f749b383e501ecbda0c43f0bbe5d4a3ffb2a28`) for the test.
- Baseline probe against the previous binary reproduced the issue: `pname=worker-thread`.
- This kernel cannot execute the argv path: direct verifier probes reject `bpf_get_current_task`, `bpf_get_current_task_btf`, `bpf_probe_read_kernel`, and `bpf_probe_read_user_str` for both `BPF_PROG_TYPE_CGROUP_SOCK` and `BPF_PROG_TYPE_CGROUP_SOCK_ADDR` with `program of this type cannot use helper`. The PR therefore took its intentional thread-`comm` fallback and the service remained startable. This host is a compatibility/fallback result, not a positive argv-capture result.
- Restored the original binary (`sha256 12e2bba768ea7fcf38fc4f35f57a94a7c5671a76a7f0453cab98a659a34b4a9c`), confirmed `/etc/init.d/honk status` is `running`, and removed all temporary artifacts.

Glassyiris (commented):
tested

PR #70, opened 2026-08-25
fix(hysteria2): harden official transport interoperability

Glassyiris:
## Summary

- harden Hysteria2 interoperability with the official v2.12.2 server: encode `Hysteria-CC-RX` in bytes/s, accept quic-go's two-segment initial flight, and reject malformed UDP fragments, invalid pins, invalid hopping ranges, port 0, and undersized Salamander secrets
- remove the common Salamander hash-input allocation and reuse one sealed-packet buffer per socket while retaining the scalar socket path
- keep the official 5-second minimum hop interval, lazy TCP setup, connection reuse, MTU/PMTUD/window controls, and fail-closed authentication behavior
- enable bounded UDP GSO only for plain QUIC paths with an explicit payload MTU above the conservative 1252-byte default; cap batches at 16 segments and retain `HONK_QUIC_GSO=0|1`

## Verification

- `ci/outbound-ci.sh`: **ALL GREEN**; 540 `honk-outbound` unit tests plus config/share-link suites and clippy
- focused local interoperability: TCP/domain TCP, UDP, 4096-byte fragmented UDP, Salamander TCP/UDP, connection reuse, wrong password, wrong obfs password, UDP-disabled refusal, invalid pin, and bounded-GSO policy
- official remote Hysteria v2.12.2: ACME PKI + pin, self-signed insecure + pin, Salamander, password failures, up/down bandwidth hints, UDP-disabled refusal, MTU 1200 with PMTUD disabled, custom receive windows, 3000/4000-byte UDP, and individually reachable hopping ports
- real gateway `10.10.10.50`: HTTP 204 in 0.111 s, 20/20 UDP echoes, Clash group delay 107 ms, zero outbound errors, one retained Hysteria2 session
- reviewer pass: no P0-P3 findings

Sanitized feature evidence and full benchmark artifacts are retained on `bench`: https://github.com/daeuniverse/honk/tree/bench/bench/results/quic-gso-2026-08-25

## Benchmark reading

On x86, the same final binary measured HY2 TCP at 6479 Mbps with GSO disabled and 6713-6845 Mbps with the 16-segment cap (+3.6-5.6%); direct stayed at 9401-9406 Mbps. TUIC TCP stayed within run variance, so this PR makes no TUIC TCP or Juicity speedup claim. RSS returned to the documented 48/43 MiB HY2/TUIC baseline in the repeat arm. ARM exploratory toggles were neutral/noisy and are explicitly not treated as final cap-16 evidence.

Bilingual analysis: https://github.com/daeuniverse/honk/blob/bench/doc/benchmark.en.md and https://github.com/daeuniverse/honk/blob/bench/doc/benchmark.zh.md

## Verified limits

- the official server receives a 4096-byte UDP request but cannot return the same payload because its 4096-byte message buffer must also contain the Hysteria header; the Rust fixture completes the full 4096-byte round trip
- public-path port migration was lossy for both honk and the official client, so no hopping reliability/performance claim is made
- the throughput harness is plain HY2/TUIC; Salamander and hopping are covered by functional interoperability tests, not those throughput rows

### DIFF
--- a/crates/honk-outbound/src/proxy/hysteria2/mod.rs
+++ b/crates/honk-outbound/src/proxy/hysteria2/mod.rs
@@ -26,13 +26,11 @@
 //! - **Salamander obfuscation** (`salamander.go`): when `hy2_obfs` is set,
 //!   every UDP datagram on the wire gets an 8-byte random salt prefix and the
 //!   payload XORed with `BLAKE2b-256(password ++ salt)` repeated. Implemented
-//!   as a custom quinn `AsyncUdpSocket` (`SalamanderSocket`).
+//!   as a custom quinn `AsyncUdpSocket` (`Hy2UdpSocket`).
 //! - **Congestion control**: without bandwidth hints the connection runs BBR
-//!   and sends `Hysteria-CC-RX: 0` (sing-quic's non-brutal default,
-//!   `client.go:580-588`). When `hy2_up_mbps` is set the send side uses a
-//!   fixed-rate brutal sender ([`crate::quic::BrutalConfig`]); when
-//!   `hy2_down_mbps` is set it is advertised via `Hysteria-CC-RX` so the
-//!   server's brutal sender paces the downlink.
+//!   and sends `Hysteria-CC-RX: 0`. `hy2_up_mbps` selects the fixed-rate
+//!   Brutal sender. `hy2_down_mbps` is serialized as bytes/s in
+//!   `Hysteria-CC-RX` so the server can pace its sender.
 //!
 //! ## HTTP/3 layer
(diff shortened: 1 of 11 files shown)


Glassyiris:
Deployment note: the development musl candidate `dadefe6873c5f6f2a75f455a73fe09984230abc9a8fe1c4f8b296ca1d9a5e83a` was manually copied to `10.10.10.50:/usr/local/bin/honk-core` and the service restarted for the TCP/UDP smoke test. It is not a release artifact. The prior binary is retained there as `/usr/local/bin/honk-core.pre-cap16`.

PR #71, opened 2026-08-25
feat(score): improve evidence-driven selection

Glassyiris:
## Summary

- replace least-attempt periodic exploration with the non-incumbent's Beta reliability upper-confidence bound while preserving deterministic cold exploration and the existing 16-64 selection cadence
- keep reliability as the admission gate, then normalize latency against the fastest close-reliability candidate and dominant-direction bytes/s against the highest rate in that band
- scale incumbent hysteresis from zero to the full `0.01` margin over eight distinct effective completions; exact/family/global copies no longer multiply the evidence
- add aggregate `switchFlap` TCP/UDP counters for committed A→B→A reversals within eight selections; intentional exploration is excluded and no node/target identity is exported
- add a 64-member Score selection benchmark and update the English/Chinese design, group, and API references

## Verification

- `ci/outbound-ci.sh`: **ALL GREEN**; 546 `honk-outbound` tests plus config/share-link suites and all-target clippy
- focused Score suite: 61 tests, including UCB orientation, scale-invariant performance, distinct reporter evidence, flap integration/window, reload pruning, and Peek/cadence behavior
- focused Clash `/stats.score` contract test: pass
- workspace all-target clippy: pass
- independent review found a layered-completion hysteresis bug; the final commit uses one most-specific non-empty evidence layer and the reviewer verified the correction with no remaining blocker

## Profile decision

`group_selection/score_peek_64` measured **5.50 µs** median. The allocation probe measured 136 allocations / 15,260 bytes per rank across the complete public selection path; the snapshot `Vec` itself is one of those allocations. Removing only that allocation is not justified, so this PR leaves the existing single Mutex and snapshot allocation intact rather than adding a reusable-buffer lifecycle.

## VyOS real-traffic test

Manually deployed musl candidate `d485e82aafc96b7f18b391dc397be4c91dbb09913b4e2e6b20e5efd359ab9380` to `vyos@10.10.10.1:/config/vyos-scripts/podman/dae/dae` with `sudo`; beta.60 is retained as `dae.pre-score-policy-9583fc62`.

- `/stats.score`: all six live Score groups expose the eight counters, including `switchFlap`
- real group-delay probes plus 112 HTTPS flows: initial cold run 111/112 (one `URLError`), warmed run 112/112, p50 171 ms / p95 691 ms
- live telemetry observed 2,163 periodic selections, 55 switch flaps, and 8 fresh-failure bypasses while background traffic continued
- `dae.service` remained active and its error journal was empty

## API

`score.groups[].{tcp,udp}.switchFlap` is additive. `/stats.score` still exposes only group names and aggregate counters; scorer cells, node IDs/tags, exact targets, family scopes, and switch history remain private and memory-only.


Glassyiris (commented):
review & tested

PR #86, opened 2026-08-28
chore(llm): add contributing rule

Glassyiris:


jschwinger233 (approved):

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+
+This policy exists so that contributors know the rules before they write code, and so reviewers can point to a public, actionable reason when enforcing them. It is not a stance on whether LLMs are good or bad; it is a statement of where understanding must live in this project.
+
+---
+
+## 1. Core principle
+
+> LLMs may **fill**; humans must **own the invariants**.
+
+It is fine to use LLMs to answer questions, analyze, distill, refine, check, suggest, review, and to generate code in the areas listed in §3. It is not acceptable for an LLM to originate changes in the soundness-critical areas listed in §2 unless the human author is already a domain expert who fully understands and stands behind every line.
```
olicesx: 🟡 **major** — These sections give opposite answers for the same contribution. This paragraph allows an LLM-originated change in a §2 area when the human author is a domain expert and stands behind every line. Sections 2 and 8 say the same change must be human-originated and must not be created by an LLM, without preserving that exception. Since the policy also allows non-compliant PRs to be closed without further discussion, contributors and reviewers do not have a single rule they can apply consistently.

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+
+The project cares more about authors **understanding** what the code does, **planning** how it will change, and **deciding** what it should look like, than about the code itself. A polished PR is not evidence of any of these.
+
+## 2. Soundness-critical areas (human-authored only)
+
+The following carry honk's core invariants — generation ownership, lock ordering, eBPF two-phase commit semantics, failure-direction guarantees. Changes here must be originated by a human who can defend the design line by line. LLMs may analyze, review, and suggest, but must not create the change:
+
+- `crates/honk-core/src/ebpf/` (real backend, map lifecycle, routing push plans)
```
olicesx, line 22: 🟡 **major** — This boundary covers the userspace eBPF backend, but not the actual kernel program in `crates/honk-ebpf/` or the shared kernel/userspace ABI in `crates/honk-ebpf-common/`. `AGENTS.md:419,502` identifies stable layouts and token semantics across these crates, and `.github/CODEOWNERS:2-5` calls them the repository's most fragile invariant. Under §3, an LLM-originated safe-code change in either omitted crate would only require human review, which is weaker than the human-origination requirement applied to the userspace backend.

olicesx (changes requested):
The policy has a clear goal, and the repository-specific guidance is useful. I found two gaps in the central rules that would lead to inconsistent enforcement: the exception in §1 conflicts with the absolute prohibition in §2/§8, and the critical-path boundary leaves out the kernel and shared-ABI crates.

🔴 CHANGES REQUESTED — 0 blocking / 2 major

<details>
<summary>📋 Review index and additional notes</summary>

- 🟡 major — The domain-expert exception conflicts with the absolute human-origination rule (`CONTRIBUTING.md:13`; see inline comment).
- 🟡 major — The critical-path boundary omits the kernel eBPF and shared-ABI crates (`CONTRIBUTING.md:21`; see inline comment).
- 🔵 minor — “Full test pass” does not currently have a consistent executable meaning (`CONTRIBUTING.md:90`). Both test jobs in `.github/workflows/ci.yml:55,100` and the current validation guidance in `AGENTS.md:375` explicitly skip `test_routing_with_config_dae`. A contributor can therefore pass every supported gate while still failing the literal wording of this requirement.
- 🔵 minor — The repository and its GitHub community profile do not currently provide a Code of Conduct (`CONTRIBUTING.md:95`). This classifies behavior under a governance document that contributors cannot read or use to find the relevant reporting and enforcement process.
- Verification: reviewed head `2ce13dd4c23d3a9665ef0b5b3fcc323fdf55da5e`; all four GitHub CI checks are green. I did not complete a local checkout because the clone attempts failed with network errors; this PR changes documentation only.
</details>

CONTRIBUTING.md
Glassyiris: 已经添加 ABI 等存在的 honk-ebpf-common

CONTRIBUTING.md
Glassyiris: 保留对领域专家的自主权，逻辑上AI会受到下部的约束。

CONTRIBUTING.md
```diff
@@ -0,0 +1,104 @@
...
+
+This policy exists so that contributors know the rules before they write code, and so reviewers can point to a public, actionable reason when enforcing them. It is not a stance on whether LLMs are good or bad; it is a statement of where understanding must live in this project.
+
+---
+
+## 1. Core principle
+
+> LLMs may **fill**; humans must **own the invariants**.
+
+It is fine to use LLMs to answer questions, analyze, distill, refine, check, suggest, review, and to generate code in the areas listed in §3. It is not acceptable for an LLM to originate changes in the soundness-critical areas listed in §2 unless the human author is already a domain expert who fully understands and stands behind every line.
```
olicesx: §1 allows an LLM to originate a §2 change when the author is already a domain expert and stands behind every line, but §2 and §8 say those changes must be human-originated, with no pointer back to the exception. §4.5 also lets us close with no questions asked.

I don't need a definition of “domain expert” — just one outcome for the same PR. Dropping the exception, or having §2/§8 say the §1 exception still applies, would both work.

CONTRIBUTING.md
```diff
@@ -0,0 +1,104 @@
...
+
+The project cares more about authors **understanding** what the code does, **planning** how it will change, and **deciding** what it should look like, than about the code itself. A polished PR is not evidence of any of these.
+
+## 2. Soundness-critical areas (human-authored only)
+
+The following carry honk's core invariants — generation ownership, lock ordering, eBPF two-phase commit semantics, failure-direction guarantees. Changes here must be originated by a human who can defend the design line by line. LLMs may analyze, review, and suggest, but must not create the change:
+
+- `crates/honk-core/src/ebpf/` (real backend, map lifecycle, routing push plans)
+- `crates/honk-ebpf-common/` (ABI， aya Pods)
```
olicesx: Thanks for adding `crates/honk-ebpf-common/`.

The kernel crate still isn't in this list. CODEOWNERS puts `crates/honk-ebpf/` first among paths that must not merge on bot approval alone, and AGENTS.md already treats it as moving with the ABI crate and the userspace backend.

§3 welcomes mechanical refactors outside §2, so as written an LLM could originate a safe refactor in the kernel tree while the sibling paths are human-only. If the kernel crate belongs here, a one-line path would make that obvious.

olicesx (changes requested):
Thanks for adding `honk-ebpf-common`. Two spots still leave me unsure how to apply this document when closing a PR — notes on the exception and the kernel crate path below.

PR #88, opened 2026-08-28
refactor(config): split node fields by protocol

Glassyiris:
## Summary

- replace the flat `Node` protocol field set with a typed `OutboundConfig` and per-protocol configuration structs
- preserve the legacy flat TOML/YAML/JSON key shape, field order, unrelated-field tolerance, and derived node identities through a dedicated serde adapter and UUID golden matrix
- migrate config parsers, outbound/core/tool consumers, examples, tests, and benchmarks to typed protocol access
- return an internal configuration error instead of panicking when a non-TLS node reaches the TLS connector boundary

## Compatibility notes

- Structured TOML/YAML/JSON keeps the legacy behavior: fields owned by the selected protocol are parsed, while fields left over from other protocols are ignored. For example, `protocol: ss` with `tls: true` remains accepted without enabling TLS.
- `store_subscribe` remains safe: it stores the raw body only after a successful parse, and a rejected refresh preserves the last valid body. `subscription_store_recovers_last_valid_fetch` covers restore after a failed refresh.
- Serialized nodes remain semantically round-trip safe and keep stable IDs, but are not byte-identical to older output for every protocol: TUIC/Juicity now mirror credentials into both generic and protocol-specific keys; Trojan/AnyTLS credentials are mirrored into `username` as well. Old/new readers remain compatible.
- Clash Hysteria2 import now reads the non-standard `auth` key before the standard `password` key. Only YAML that supplies both with different values changes the effective credential and derived ID.
- Fixed UUID goldens now cover SS, SOCKS5, Trojan, VMess, all six VLESS modes, Hysteria2, TUIC, Juicity, and AnyTLS.

## Validation

- `just outbound-ci` (`honk-config`: 164 passed; `honk-outbound`: 570 passed; fmt and clippy clean)
- `cargo clippy -p honk-core --bench reload --no-default-features --features reload-bench-counters,reload-alloc-bench -- -D warnings`
- `HONK_RELOAD_METRICS_ONLY=1 cargo bench -p honk-core --bench reload --no-default-features --features reload-bench-counters,reload-alloc-bench -- --noplot` (`RELOAD_METRICS` emitted)
- `cargo test -p honk-core --lib subscription::tests::subscription_store_recovers_last_valid_fetch -- --exact`
- `env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY cargo test --workspace --no-fail-fast -- --skip test_routing_with_config_dae` (1826 passed, 4 ignored, 1 filtered)

### Known routing-test exclusion

`test_routing_with_config_dae` was also run with `/etc/dae/geoip.dat` and `/etc/dae/geosite.dat` present. It reaches the standing legacy failure: the test expects `test.jogiyw.sbs:443 -> direct`, while the current `config.dae` contains no matching direct suffix rule, so the generic port rule returns `omg`. This PR changes neither `config.dae` nor that test. The repository CI and release workflows already exclude this stale assertion; that is the reason for the `--skip` above.


Glassyiris (commented):
reviewed in local

PR #93, opened 2026-08-29
feat(group): back off score exploration for failing leaves

Glassyiris:
## Motivation

Score's failure evidence decays with a fixed 30-minute half-life — including the completion counters that decide whether a leaf is "cold". A permanently dead leaf therefore turns cold again every few half-lives, and cold exploration (least-attempts-first) picks it right back: on a gateway whose subscription contains dead nodes this produced ~26k dial-timeout/DoH-failure log lines per day, with each re-exploration costing one real user flow a 9s dial timeout.

## Change

Track a consecutive-failure streak and an `explore_not_before` instant per score cell, **outside** the decaying evidence:

- each failure backs the leaf out of cold and periodic exploration for 5 minutes, doubling per consecutive failure, capped at 6 hours
- a success restores exploration eligibility immediately, but only steps the streak down by one — a flapping leaf cannot wash out its backoff history with one lucky probe
- reliability ranking, hysteresis, and the reason counters are untouched; backed-off leaves simply fall through to normal ranking (where their decayed/trained-failure reliability already loses)

No new configuration knobs, consistent with Score's existing non-configurable design.

## Tests

- `cold_exploration_skips_backed_off_candidates` — unit-level `best_index` filter
- `exploration_backoff_grows_with_failures_and_shrinks_with_success` — backoff expiry, doubling, and step-down-on-success with injected time

`cargo test -p honk-outbound group::score` (64) and `group::` (123) pass; clippy `-D warnings` and fmt clean. Docs updated in AGENTS.md and both design trees.

---

## Also in this PR

`fix(control): skip score QUIC probe when the node's UDP probe failed` — the periodic DataUdp score probe opened a doomed TLS-in-QUIC handshake through nodes whose own DNS UDP probe had just failed; each one died as a quinn endpoint-driver `I/O error` ERROR log (~450/day on a gateway with dead subscription nodes). The failure evidence was already recorded by the health probe, so the handshake is now skipped until the node passes again.


Glassyiris (commented):
tested

PR #96, opened 2026-08-29
fix(anytls): detect silent session death and bound writer queues

Glassyiris:
## Motivation

Production incident: a flapping AnyTLS server (accepts connections in bursts, then goes silent for seconds) wedged honk's session pools. The dead session never errored — TCP stayed open, no frames arrived, streams never released — so the pool's capacity wait blocked new dials into a 9s timeout for hours until a manual restart. Verified live: 0/374 engine dials succeeded while fresh single-shot clients got through 2/3 of the time.

## Changes

- **SYNACK liveness proof** (v2 peers, negotiated via `CMD_SERVER_SETTINGS`): every stream open on a reused session arms a 3s watchdog; any inbound SYNACK disarms it. A server that has gone silently dead fails the session within 3s of the next open, and the pool rebuilds. Idle dead sessions cost nothing until used — detection happens on use, not on a heartbeat timer.
- **Bounded writer queue**: the previously unbounded frame queue is capped (with reserved control headroom); overload or close makes push fail and the session terminal instead of growing memory against a wedged writer.
- Writer loop: closed-aware pop, byte-accurate batch reserve.
- `fix(group): keep setup-only probes usefulness-neutral` — setup-only health/urltest warm-ups no longer count as useful successes or failures in Score evidence.

## Tests

- `reused_v2_session_requires_synack_within_deadline` — SYNACK keeps the session live, missing SYNACK kills it
- `writer_queue_bounds_control_pressure_and_rejects_after_close`

78 anytls tests pass in the worktree.

## Notes

- Requires a v2-capable server for the liveness proof (observed in production: the provider's servers send SYNACK). v1 servers keep prior behavior.
- The 3s timeout can false-positive under extreme server load; cost is one session rebuild (~300ms).


Glassyiris (commented):
tested on vyos

PR #98, opened 2026-08-29
feat(log): silence quinn endpoint driver-death noise by default

Glassyiris:
## Motivation

`quinn::endpoint` logs every endpoint-driver death at ERROR. On a gateway with AnyTLS nodes this fires constantly — probe and warm QUIC-over-UoT endpoints die whenever their session retires, and the new SYNACK watchdog (PR #96) kills silently-dead sessions *on purpose*, so the log now partly records successful self-healing. Measured: ~75 ERROR lines/minute with zero actionable content. Tonight's production analysis confirmed no functional impact: the probe records its own outcome, and warm transports re-establish on the next warm tick.

## Change

The default log filter gains `quinn::endpoint=off`. `tracing`'s EnvFilter cannot remap levels, so silencing the target is the only way to demote this noise. A valid `RUST_LOG` still wins over the default, so debugging keeps full access.

Genuine QUIC client endpoint failures (TUIC/Juicity/Hysteria2) remain visible through their dial/connection error paths; the endpoint-level line was redundant with those.


Glassyiris (commented):
using for:
I/O error: AnyTLS UoT stream is no longer registered
18714
error
05:38:13
I/O error: AnyTLS UoT stream is no longer registered

PR #102, opened 2026-08-30
feat(config): accept official hysteria2 embedded hop ports in share links

Glassyiris:
## Summary

Accepts the official Hysteria2 client form with hop ports embedded in the authority:

```
hysteria2://letmein@example.com:123,5000-6000/?insecure=1&obfs=salamander&obfs-password=gawrgura&pinSHA256=deadbeef&sni=real.example.com#cool-server
```

The full list becomes the hop set (`hy2_port_hopping`); the first entry stays in the address as the nominal port. Since `Node::derive_id` does not hash the hop list, both spellings of the same server keep identical node identity (health/latency/session state survives a subscription flip between the forms) — pinned by test.

Rules:

- `hysteria2://` and the `hysteria://` alias both supported; range-only (`:5000-6000`), mixed lists, and bracketed IPv6 work
- An embedded list fails structural validation at parse time (nonzero ports, ordered ranges, digits only) instead of failing at dial
- Specifying both the embedded form and `mport=` is rejected as ambiguous; an empty `mport=` is absent, not a conflict
- Non-hy2 schemes, plain single ports, and bare (invalid) IPv6 literals take exactly the pre-change code path

## Tests

- `test_hysteria2_embedded_hop_ports` — the full example link, range-only form, alias scheme, identity equivalence across spellings, mport conflict, five malformed lists, plain-port and IPv6 regression guards, no-suffix form, whitespace-only list, empty `mport=`
- `test_embedded_hop_spec_parity_with_share_link_parser` (honk-outbound) — pins the parse-time validator and the dial-time `parse_port_hopping` to the same accept/reject table so the two copies cannot drift apart silently
- `cargo test -p honk-config` (166) and `cargo test -p honk-outbound --lib hysteria2` (41) pass; clippy and fmt clean


Glassyiris (commented):
reviewed by @Glassyiris 

code: @Glassyiris 
test/ci/doc: kimi k3

PR #106, opened 2026-08-30
perf(score): damp hold margin on same-target winner flaps

Glassyiris:
## Problem

Production counters (`/stats.score`, 4h window) show the `america` Score group oscillating with **switchFlap = 1391** while `failStreakExcluded = 0` — reversals driven purely by ranking noise between near-equal nodes, not by failures. Each flap costs a needless outbound switch.

## Change

Adaptive per-target-scope hysteresis in the Score policy:

- Each committed winner reversal within 8 selections of the same `(group, network, family, target)` scope raises that scope's hold margin by one power of two (cap **8x** = 0.08).
- Each held selection lowers the damping by one level.
- `FreshFailureBypass` is unaffected: a failing incumbent still loses immediately regardless of damping.
- No new configuration surface; state rides the existing bounded `selection_history` LRU (4096 entries).

## Evidence

- `cargo test -p honk-outbound --lib group::score`: 71 passed — new tests `switch_flap_damps_hold_margin_per_target` (raise/cap/decay), `hold_decision_flap_margin_multiplies_switch_threshold` (multiplied threshold + fresh-failure bypass at max level), `flap_margin_reaches_hold_decision` (rank-level wiring).
- `cargo clippy -p honk-outbound --all-targets` clean; `cargo fmt --all` applied.
- Docs synced: AGENTS.md module map, `doc/en/design/groups.md`, `doc/zh/design/groups.md`.

## Notes for operators

- After this lands, part of what used to count as `performanceWinner` switches now surfaces as `freshFailureBypass`/`incumbentHeld` — do not read the shift as a failure-rate change.
- Known bound: a sustained deterministic ping-pong converges to a hold/switch limit cycle, so the flap-rate ceiling is halved rather than eliminated. If production `switchFlap` stays high, the follow-up is decaying the level only after a full window of held selections.


Glassyiris:
尝试使用这个进行抖动减少，reviwed
```
有效 margin =
  0.01
  × incumbent evidence factor
  × 2^margin_level
```

design: @Glassyiris 
tests/ci/code: kimi k3

PR #108, opened 2026-08-30
fix(group): peek unchosen score sub-groups under selector parents

Glassyiris:
## Problem

Every flow through a Selector group flattens **all** of its sub-group members with `SelectionEffects::Apply` to build the candidate list. For Score sub-groups that means an authoritative rank — selection counters, incumbent marks, and flap history — is committed for each of them, including ones the selector did not choose.

Production evidence: a Score sub-group that no traffic ever routes through still accumulated ~150 ranks/min and a growing flap count, driven entirely by sibling traffic through its Selector parents (`ai`, `country`, ...). Since the unused group has no real flows, its evidence decays cold, its hysteresis margin collapses to zero, and every candidate-resolution rank flips the winner — a self-sustaining flap generator that pollutes `/stats.score` and any damping logic built on it.

## Change

In `flatten_candidates`, a Selector parent with applied effects resolves only its **chosen** sub-group with `Apply`; all other sub-groups are resolved with `Peek`. Other policies (Score/URLTest/LoadBalance/Fallback) are unchanged — they genuinely need every member's committed pick.

Semantics preserved:

- Chosen sub-group: unchanged authoritative selection.
- Unchosen sub-groups: candidate list, alive filtering, and selector fallback still work — only selection bookkeeping is skipped. Fallback dials through a peeked leaf still feed that sub-group's Score evidence via attribution.
- Switching the selector to a previously unchosen sub-group starts it cold, exactly as before — the phantom Apply ranks never trained any evidence, so nothing of value is lost.

## Evidence

- New test `selector_parent_peeks_unchosen_score_subgroups`: dialing through a Selector parent records a rank only for the chosen Score sub-group; unchosen sub-group counters and flap history stay empty; switching the choice moves the committed rank.
- `cargo test -p honk-outbound --lib group`: 132 passed; clippy + fmt clean; `honk-core` clash API suite green.

Follow-up to #107 (same pollution family: that one covered dashboard-driven resolution, this one covers sibling-traffic-driven resolution).

## Update (second commit)

Production verification with a temporary rank-caller log showed the first commit alone did not stop the phantom ranks: the **target-aware dial path** (`flatten_candidates_for_target` in `score.rs`, used by `selection_plan_for_target*`) is a separate flattening that applied effects to every sub-group unconditionally. The same chosen-only rule is now applied there; the test covers both paths.

Post-fix production counters (70s window): previously-phantom sub-groups (`hongkong`/`japan`/`america`) record zero ranks; only genuinely chosen groups rank (`wind` 264, `sgp` 100, `taiwan` 12 — matching the live selector choices).


Glassyiris:
减少结构性的 group in group 中存在的过度探测

reviewed by @Glassyiris 
code: @Glassyiris 
test/ci: Kimi K3

Glassyiris:
Follow-up triage:

1. **URLTest/Fallback parents** — valid, same shape, zero production trigger today (no URLTest/Fallback parent with Score sub-groups in observed configs). Filed as #109 with the chosen-tag-equivalent plan (urltest_cache incumbent / fallback pin).
2. **Target-aware flatten** — already covered: the second commit (116442d) adds the same chosen-tag rule to `flatten_candidates_for_target`, recursively, so Score→Selector→Score nesting is handled. (Reviewed head may have been 204d5f5.)
3. **Stale chosen_tag** — fixed in 68f531d: `selector_chosen_tag` filters the stored choice against current members and falls back to the first member, so the actually-serving sub-group keeps committing ranks.

Glassyiris:
阻塞项修复（head 18c19af):

两处都成立，根因是 chosen_tag 在 flatten 阶段预判服务者，与 pick_selector 的真实逻辑（valid stored → valid default → first alive）必然有分歧面。已放弃预判，改为 **commit-after-pick** 架构:

- Selector 在 flatten 阶段一律 Peek 全部子组
- pick 完成后，对实际服务子组用 Apply 重新解析提交（无论服务路径是 stored choice / default / alive 回退，提交永远跟随真实服务）
- 四个选择点（domain 计划、嵌套解析、target 计划、target 嵌套）共享 `commit_selector_pick` / `commit_selector_pick_for_target` 两个 helper
- 回归测试：`selector_commit_follows_non_first_default`、`selector_commit_follows_alive_fallback` 覆盖你复现的两个场景

135 group 测试全绿。

Glassyiris (commented):
引入对 selector -> score 这种情况下的
**stale choice + 合法非首位 default 仍会把 Score 副作用记错子组**
的问题

PR #112, opened 2026-09-01
fix(outbound): stop session-pool stalls behind dial backoff and speculative capacity

Glassyiris:
## Symptom

User report: healthy AnyTLS nodes intermittently fail every dial with `dial timed out after 9s` while HTTP health checks through the same node succeed in ~230ms. Reproduced live on the production gateway with a large AnyTLS subscription whose servers kill some fresh sessions instantly.

## Root cause (one defect chain in `SessionPool`)

A server that accepts the dial but kills the session immediately ("dead on arrival") drives the pool into a state where every later flow parks until its outer dial deadline (`connect_timeout * 3` = 9s) kills it — with zero dial activity in between:

1. **Dial "success" hid the failure from the breaker.** The dial task completes (TCP+TLS+auth all fine), pushes the session, and resets `dial_failures`; the session dies microseconds later. The breaker never engages, so the pool hot-spins fresh dials every ~170ms (instrumented on the gateway) and waiters that keep losing the re-registration race ride `Step::Wait` until their 9s budget expires.
2. **The dead-on-arrival error path left a phantom in-flight dial.** When the registering caller loops back, finds the fresh session already pruned, and errors out ("session established but immediately unusable"), it had *already* published `pool.dial_done = Some((id, Pending))` for its second registration and never spawned the task. No task, no `DialGuard`, no cleanup: every subsequent `offer()` parks in `Step::Wait` on a channel nobody will ever signal. The wedge is permanent until restart — and health checks never see it because cold reusable nodes are probed through guarded ephemeral pools (`probers.rs`). This was the dominant production failure: multi-minute storms of 9s timeouts per node while probes kept reporting the node alive.
3. **Backoff parks amplified everything.** `Step::Backoff` slept the caller for the whole remaining window (up to the 30s `max_dial_backoff` cap), so once real dial failures piled up, flows arriving inside the window died at 9s without a dial being attempted, and the artificial timeouts fed back into dial-failure strikes.
4. **Speculative provisional slots counted against normal offers.** Hung cold-race UDP checkouts could push a cold pool (`max_sessions: 2`) into `Step::Capacity`, another silent unbounded park.

## Fixes

- `fix(outbound): cap session dial backoff below the flow dial budget` — `max_dial_backoff` 30s → 2s; the breaker only paces redials, single-flight already paces concurrent dials.
- `fix(outbound): keep speculative slots from starving normal dials` — `offer()` bounds the normal path by real Active sessions only; `checkout_speculative()` keeps counting provisional slots for its own discipline.
- `fix(outbound): unwedge session pools behind dead-on-arrival sessions` — the dead-on-arrival error path now clears its phantom `dial_done` entry (dropping the sender wakes waiters to re-elect) and records a dial failure so the breaker paces redials; `Step::Backoff` in both `offer()` and `checkout_speculative()` fails fast with the recorded error instead of parking, so flows get a real error in milliseconds and client retries pick a recovered server up within the 2s cap.

## Tests

- New: `dead_on_arrival_session_engages_the_breaker` — reproduces the production wedge: the pre-fix code hung this test forever (phantom `dial_done` parked the second offer); post-fix it errors fast, arms the breaker, and redials after the window.
- New: `dial_backoff_is_capped` — escalation respects `max_dial_backoff`.
- Rewritten: `dial_failures_back_off` (fast-fail inside the window + paced redial after it), `provisional_slot_blocks_normal_offer_until_released` → `provisional_slot_does_not_block_normal_offer` (the old test pinned the wedging behavior).
- Session suite 29/29, AnyTLS suite 81/81, `ci/outbound-ci.sh` all green.

## Production validation

Deployed on the gateway (real eBPF, nexi AnyTLS subscription). Pre-fix: storms of `dial timed out after 9s` per node with zero pool-side logs in between (82 events / 15 min), matching the user's report. The wedge was reproduced and pinned by the new unit test (the pre-fix code hangs it forever).

Post-fix, 15-minute window under normal LAN load: **zero** 9s dial timeouts, 2 dead-on-arrival events resolved as fast errors, 19 real `session dial failed` fast errors, backoff never escalated past 1s, 527 healthy proxied TCP connections.

### DIFF
--- a/crates/honk-outbound/src/session.rs
+++ b/crates/honk-outbound/src/session.rs
@@ -16,9 +16,11 @@
 //!   drops and only an explicit winner commit may publish; provisional slots
 //!   bound speculative work only — normal offers are bounded by real sessions,
 //!   so hung speculative dials can never starve them;
-//! - dial circuit breaker: consecutive establishment failures back off
-//!   exponentially before the pool dials again (a dead server must not
-//!   eat a TCP connect per proxied flow);
+//! - dial circuit breaker: consecutive establishment failures (including
+//!   dead-on-arrival sessions) pace redials with a capped backoff, and
+//!   callers fail fast inside the window instead of parking — a dead server
+//!   neither eats a TCP connect per proxied flow nor stalls flows until
+//!   their outer dial deadline;
 //! - RAII stream-slot permits (`SessionPermit`) as the single capacity
 //!   truth, and [`SessionPool::open_with`] for atomic reserve+open;
 //! - idle reaping, jittered max-age drains and optional prewarm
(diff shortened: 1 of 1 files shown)


Glassyiris:
reviewed by: @Glassyiris 
code/detect: @Glassyiris 

test/ci: Kimi K3

Issue #113, opened 2026-09-01
AnyTLS probe churn may trip provider per-account session rate limits


Glassyiris:
## Background

While investigating #112 (session-pool wedge, fixed in v0.0.1.beta.67) we captured the triggering server behavior on the production gateway (nexi AnyTLS subscription): the server accepts TCP+TLS+auth and then **closes the session with a clean FIN ~50–400ms after establishment — no AnyTLS `ALERT`, no read error**. Kills are intermittent and cluster in time (several nodes killed simultaneously for minutes, then it stops).

## Evidence

- All ~55 nexi nodes share one account credential (same password across hosts).
- honk's session establishment rate against the fleet: **sustained ~2.75 new TLS sessions/s, bursts to 6/s** (1,651 sessions / 10 min). The bulk is health-check churn: a cold node gets a fresh ephemeral session per probe (one exchange, then close), every 30s cycle, TCP + UDP.
- Zero-stream session closes are rare at idle (8/1,581 = 0.5% over 10 min) but spike during storm windows.
- Reference clients (mihomo/sing-box) keep **one long-lived multiplexed session per node** and never trip this — matching user reports that the same nodes work fine there.

Hypothesis: the provider enforces a per-account (or per-account-per-IP) limit on **new session rate / concurrent sessions**, and honk's probe-driven churn trips it periodically. The kills stop being user-visible after #112 (fast error + breaker instead of a pool wedge), but the provocation remains.

## Proposed directions

1. **Probe session reuse**: keep one persistent per-node "health" session for AnyTLS (and other muxed protocols) that probes share, instead of an ephemeral session per cold-node probe. Trade-off: one long-lived connection per node vs. session-establishment churn.
2. **Probe cadence for healthy cold nodes**: back off probing for nodes that are alive but unselected, not just for dead ones.
3. **Confirm the limiter**: controlled A/B — open sessions from a reference client (sing-box) to the same nodes at low rate vs. honk's probe rate and compare kill rates.

## Acceptance criteria

- Session-establishment rate per node/account drops by an order of magnitude under the default health-check configuration.
- No regression in dead-node detection latency (recovery scheduler behavior unchanged).
- Documented guidance for subscriptions whose providers rate-limit session establishment.

Related: #112

PR #116, opened 2026-09-01
fix(anytls): reset only the unanswered stream when the session stayed active

Glassyiris:
## Symptom

Production gateway, after the #112 fixes: bursts of relay errors like

```text
TCP relay error for 10.10.10.90:64971 → 149.154.171.255:80: stream 17 SYNACK timed out after 3s
```

six+ flows failing at the same millisecond with the same stream id. Forensics on one burst: session `8093` carried 13 live streams (Telegram/Cloudflare targets) when one newly opened stream (sid 17, chatgpt.com) got no SYNACK within the 3s deadline; the watchdog failed the **whole session**, erroring every sibling flow. The server was healthy — it kept the session up and was serving other streams — it just never answered that one open (most likely its own dial to the target stalled).

## Root cause

The per-SID SYNACK watchdog (#96/#97) treats every unanswered open as a silently-dead session. That conflates two different conditions:

- **Silent session death** — no inbound frames at all during the window: the kill is correct and is what the watchdog exists for.
- **Live session, unanswered open** — inbound frames kept flowing while one open went unacknowledged: the server is slow to open that stream (its target dial is not protocol-bounded), and killing the session converts one stream's problem into an outage for every sibling.

The reference sing-anytls effectively tolerates the second case on busy sessions: its single deadline watcher is cancelled by *any* SYNACK, so a constant stream of acknowledgements keeps it from firing. honk's per-SID tracking (deliberate — the reference shape loses deadlines of concurrent opens) removed that tolerance entirely; this restores it precisely, keyed on session activity rather than on unrelated SYNACKs.

## Fix

The session counts inbound frames (`rx_frame_seq`, bumped per demux frame). The deadline task snapshots the counter when the SYN hits the wire; on expiry:

- frames arrived during the window → **reset only that stream** (`dispatch_error`, the flow retries while siblings keep flowing);
- zero frames during the window → fail the session as before (true silent death).

## Tests

- New: `synack_timeout_on_active_session_resets_only_the_stream` — pins the production case (active session + unanswered open → stream error, session and sibling stream survive).
- Updated: `synack_deadline_is_tracked_per_stream` and `synack_deadlines_do_not_share_elapsed_time` — per-SID firing semantics unchanged; the consequence on an active session is now the stream reset instead of session death.
- Unchanged and still green: the silent-window case (`reused_v2_session_requires_synack_within_deadline` — no frames, session dies).
- AnyTLS suite + `ci/outbound-ci.sh` all green.

## Production validation

Gateway, 28-minute window under normal LAN load, pre-fix vs post-fix:

- pre-fix: 171 session-killing SYNACK timeouts / 2h, each burst erroring every sibling flow on the session (13 streams in the analyzed case);
- post-fix: 8 timeouts / 28 min, all on one quiet session with no inbound frames in the window (legitimate silent-death kills), zero 9s dial timeouts, 2,231 healthy proxied TCP connections.

Related: #112, #113, #114


Glassyiris (commented):
reviewed

PR #121, opened 2026-09-02
fix(routing): prevent LPM exhaustion during reload

Glassyiris:
## Summary

- prune LPM keys unused by either routing generation before staging replacement values
- restore the LPM trie capacity to the shared 2,048,000-entry limit for large GeoIP sets
- cover failed replacement writes and active-plan replay with a regression test

## Tests

- `cargo test -p honk-core --lib control::routing_matcher::tests` (40 passed)
- `cargo test -p honk-core --lib routing_push_failure_replays_old_plan_and_keeps_userspace_generation`
- `cargo +nightly build --release -Zbuild-std=core --target bpfel-unknown-none` in `crates/honk-ebpf`


Glassyiris (commented):
reviewed by @Glassyiris 

only pr body write by ai

PR #123, opened 2026-09-02
feat(config): support subscription user agents in dae

Glassyiris:
## Summary

- support `tag: 'url'(user-agent)` subscription entries
- support structured `tag: { url, ua, interval }` entries using the existing subscription runtime fields
- preserve legacy bare URLs containing parentheses or fragments and document both forms in English and Chinese

## Testing

- `cargo test -p honk-config`
- `cargo test -p honk-core subscription::tests --lib`
- `env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY cargo test --workspace --no-fail-fast -- --skip test_routing_with_config_dae`
- `cargo fmt --all -- --check`


Glassyiris (commented):
reviewed

PR #125, opened 2026-09-03
fix: review follow-ups for the #111–#124 batch

Glassyiris:
Follow-ups to the post-#110 review of the #111–#124 batch. 

- `fix(socks5)`: the TCP handshake encoded the domain length with `as u8`; a >255-byte domain wrapped the length and desynced the request. The UDP path already used `u8::try_from`; the TCP path now matches it.
- `fix(outbound)`: normal offers deliberately ignore provisional slots, so a speculative `commit()` could publish into a pool that had meanwhile filled past `max_sessions`. `commit()` now re-arbitrates under the lock: at capacity the winner is admitted drain-only (`begin_drain`), which is safe because speculative winners take their own stream permit before committing.
- `fix(core)`: subscription rebase on reload matched only by URL while the cache identity is URL + UA + headers, so two same-URL subscriptions with different user agents could swap IDs, carried-over nodes, and `subtag()` attribution when reordered. Rebase now matches on the same full fetch identity.
- `fix(config)`: `Config::validate` never looked at subscriptions; it now rejects an empty name, an empty URL, and non-http(s) URL schemes.

Deferred from the same review: the AnyTLS flush/shutdown completion barrier (writer-queue drain semantics — a design change, not a one-liner) and the QUIC/LPM memory-budget proposal.

Tests: new focused tests for all four fixes; `cargo test --workspace --no-fail-fast -- --skip test_routing_with_config_dae` and `cargo clippy -p honk-config -p honk-outbound -p honk-core --all-targets -- -D warnings` are green.

### DIFF
--- a/crates/honk-config/src/config.rs
+++ b/crates/honk-config/src/config.rs
@@ -760,6 +760,26 @@ impl Config {
             check_outbound(rule.outbound.as_str(), false)?;
         }
         check_outbound(&self.routing.default_outbound, true)?;
+        for subscription in &self.subscriptions {
+            if subscription.name.is_empty() {
+                return Err(crate::ConfigError::Validation(
+                    "subscription name must not be empty".into(),
+                ));
+            }
+            if subscription.url.is_empty() {
+                return Err(crate::ConfigError::Validation(format!(
+                    "subscription '{}' has an empty url",
+                    subscription.name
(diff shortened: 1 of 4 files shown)


Glassyiris (commented):
reviewed

PR #127, opened 2026-09-03
fix(core): restore datapath admission after a rejected NFQUEUE fence

Glassyiris:
Production incident: after an overnight uptime, a reload (SIGHUP or subscription merge) hit `failed to fence NFQUEUE before reload: failed to quiesce staged UDP decisions`, the reload was rejected — and the engine then rejected **every** new connection forever (direct, proxy, DNS all timing out) while health probes kept working. Only a restart recovered.

Root cause of the zombie: the fence-failure branch in `apply_runtime_config_locked` armed `datapath_healthy = false` plus both drain trackers and returned `false` without restoring anything. Nothing was torn down at that point — the fence runs before any drain — so the old generation was perfectly serviceable, yet admission stayed closed permanently (`datapath_healthy=false` also gates every later reload attempt).

Fix: the fence-failure branch now drains pending UDP admission and restores the old datapath flags + reopens NFQUEUE via the same `restore_datapath_flags_after_rejected_reload` path the sibling rejection branches already use, leaving the datapath healthy and admitting on the old config. A persistent quiesce failure now degrades to "reloads rejected, traffic unaffected" instead of a dead gateway. The error log also carries the full anyhow chain (`{error:#}`) so the inner quiesce cause (grace-period timeout vs token mismatch) is visible next time.

Regression test: `fence_failure_rejects_reload_without_stranding_datapath` injects a fence-stage flags-write failure and asserts the reload is rejected, the old config is retained, the datapath stays healthy, neither drain tracker rejects, and the write trace is exactly fence(failed) → set_static → reopen.

The deeper question — *why* quiesce failed after long uptime — is not answered by the surviving logs; the improved error chain is meant to capture it on recurrence.

### DIFF
--- a/crates/honk-core/src/control/reload/transaction.rs
+++ b/crates/honk-core/src/control/reload/transaction.rs
@@ -723,14 +723,14 @@ impl ControlPlane {
             pending.cancel_all().await;
         }
         if !self.udp_pool.cancel_initializers_and_wait().await {
-            warn!("UDP initializers did not drain after NFQUEUE reopen failure");
+            warn!("UDP initializers did not drain during admission teardown");
         }
         #[cfg(feature = "ebpf")]
         if let Some(pending) = self.pending_udp_verdicts.as_ref() {
             pending.wait_empty().await;
         }
         if !self.udp_pool.wait_for_retirements().await {
-            warn!("UDP endpoint retirements did not drain after NFQUEUE reopen failure");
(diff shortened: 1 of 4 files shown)


Glassyiris (commented):
reviewed

当前仅对可用性进行快速修正，具体问题需进行复现

only pr body by ai

PR #129, opened 2026-09-03
feat(outbound): measure delay as one warm-path round trip

Glassyiris:
## Background

Delay measurement previously used a cold, full-chain convention: proxy TCP connect + REALITY/TLS handshake + target TLS + HEAD, all timed. Protocols without reusable sessions (VLESS legacy/uot/xudp, trojan, vmess, ss) paid the entire handshake chain on every measurement, while warmable ones (AnyTLS, H2MUX/Mux.Cool, QUIC) only excluded session setup — two different conventions on the same panel. Real-node data (40-node VLESS reality+vision+xudp subscription):

| node | old value | actual composition |
|---|---|---|
| hk.15 | 85–102ms | TCP 9ms + reality 24ms + exchange 20ms |
| ca.38 | 1281–1360ms | TCP 185ms + reality 385ms + exchange 460ms |
| au.41 | 1536–1643ms | TCP 9ms + reality 600ms + exchange 228ms |

The high values came entirely from the handshake chain and server-side processing — not from the node's forwarding quality, which is what the panel number is supposed to rank.

## Change

**Every protocol now reports one warm-path round trip:**

- The measurement layer sends two requests on the established connection: the first (HEAD, keep-alive, untimed) absorbs all setup (proxy dial, target TLS, first exchange); the second (configured method/HEAD, `Connection: close`) is the reported sample.
- A server that closes after one response falls back to the first exchange's time, so measurability is never lost.
- Only two files change: `honk-outbound/src/urltest.rs` (`exchange_head`/`exchange_head_h2`/`measure_head_exchange`; the Hysteria2 timing-reset special case is deleted — now uniform across protocols) and `honk-core/src/control/probers.rs` (`http1_exchange`, the source of panel history, same convention).
- No protocol handler changes; failure accounting (streak, liveness, Score-neutral feedback) unchanged; the outer timeout still covers warm-up + both requests.

Same nodes under the new convention: hk 28–33ms, gb 234ms, ca 304ms, au 464ms — i.e. the real in-tunnel RTT.

## Tests

- New: `test_exchange_head_reports_warm_round_trip` (first response stalled 200ms, sample must be <100ms — pins warm-up exclusion)
- New: `test_exchange_head_falls_back_when_server_closes` (single response then close → falls back to the warm sample)
- The mock HTTP server now loops with keep-alive; existing 38 urltest tests and 30 clash API tests stay green
- Workspace suite + clippy `-D warnings` + fmt clean (two anytls/hysteria2 timing tests flaked under a parallel build's CPU contention and passed solo; no interaction with this change)
- Verified live on 10.10.10.49 (WAN mode) with the 40-node subscription, per the table above

Docs synced: AGENTS.md, doc/{en,zh}/reference/api.md, doc/{en,zh}/configuration.md.

## Semantics note

This is an intentional convention change: panel delay moves from "cold full handshake" to "warm-path 1 RTT" and is no longer comparable to mihomo/Clash cold-dial numbers (honk reports significantly lower values on the tested nodes). Every node's displayed delay will drop after upgrade — expected.


Glassyiris (commented):
reviewed by @Glassyiris 

design: @Glassyiris 
code/test/ci: Kimi K3

PR #130, opened 2026-09-04
fix: UDP health no longer dies on a blocked DNS check target; selector fallback gains visibility

Glassyiris:
## Problem

Observed on the gateway: `final` (selector → `proxy`, choice `wind`) kept sending **UDP** flows to `hy2-mami` while the dashboard showed `wind`. TCP was unaffected. The chain log showed `final → proxy → singal → hy2-mami`.

Root cause, in three layers:

1. **UDP liveness had a single point of failure.** A node's UDP health (both `DataUdp` and `DnsUdp`) hinged on one minimal DNS query to the first `udp_check_dns` target (default `8.8.8.8:53`). Relay servers commonly block outbound UDP/53 (anti-amplification), so the probe failed 3× and marked the node dead on both domains — while real QUIC traffic through the same node worked fine (17 recorded `performanceWinner` UDP selections in `wind`).
2. **Death was permanent.** Once excluded from UDP selection, a node receives no real UDP traffic, so `report_available_traffic` could never fire; the Score QUIC handshake probe (`h3` to the HTTPS check URL) succeeded but deliberately never touched liveness, and it was even skipped when the DNS probe failed. The only revival path was the probe that can never succeed.
3. **The Selector fallback was silent.** With `wind`'s members all UDP-dead, the `proxy` selector's configured choice was health-filtered out of the candidate set and traffic fell back to the first alive member (`singal` → `hy2-mami`) with no log; the dashboard `now` kept displaying `wind`.

## Changes

**`fix(alive): keep UDP data path alive when the DNS check target is blocked`**

- `UdpProber::probe_udp` now returns `UdpProbeOutcome { dns, data_path }`: the DNS exchange result plus the independent Score data-path handshake result.
- honk-core runs the Score TLS-in-QUIC handshake probe **regardless of the DNS probe outcome** (it was previously skipped on DNS failure — exactly the case where its independent evidence matters). It still runs only for nodes in Score groups (no reporter, no probe, no load change elsewhere) and remains `streak_neutral`.
- `probe_node_udp` now maps outcomes truthfully: DNS success → both UDP domains alive (unchanged); DNS failure + handshake success → only `DnsUdp` records the failure, `DataUdp` is marked alive from the handshake RTT; both failed → both dead (unchanged). The dial-failure streak is never touched by probes.

**`fix(group): warn when a health-filtered selector choice falls back`**

- `pick_selector` / `pick_direct_selector` now log a rate-limited warning (60 s per group+network) when the configured choice is health-filtered and traffic falls back to `default`/first-alive member. The stored choice is never rewritten, so recovery restores it.

## Tests

- `alive`: new `test_probe_node_udp_dns_blocked_but_data_path_ok_keeps_data_udp_alive` (DnsUdp dies after 3 failures, DataUdp alive with the handshake RTT); existing success/failure/timeout mocks migrated to `UdpProbeOutcome` with unchanged semantics.
- honk-core: `quic_probe_still_runs_when_dns_probe_fails` proves the handshake is attempted on DNS failure (2 dials, `dns` err + `data_path` attempted); `quic_failure_trains_score_without_failing_dns_udp_health` updated for the new return type.
- `group`: `selector_choice_filtered_by_health_falls_back_without_rewriting_choice` pins fallback + choice preservation.
- Full workspace suite, `clippy --all-targets -D warnings`, and `fmt` are green.

## Out of scope

- Nodes in non-Score groups still rely on the DNS probe alone (no reporter exists, so no handshake is attempted) — unchanged behavior, documented.
- Selector fallback *semantics* are unchanged (choice → default → first alive); only visibility was added.


Glassyiris (commented):
reviewed by  @Glassyiris

PR #132, opened 2026-09-04
fix(alive): never mark the direct builtin dead

Glassyiris:
## Problem

Production incident on a gateway (10.10.10.1, twice in one day): a LAN BitTorrent client hammering unreachable peers (`Direct/block dial ... failed (direct): connect timeout`, ~1800 in 25 min) accumulates **real-traffic failures against the `direct` builtin**. At the TCP traffic threshold (10) `direct` flips dead, which (a) ejects `direct` from Selector groups (a `direct` choice falls back to a proxy member) and (b) kills the group-OR connectivity slot of any group whose remaining members are all dead, so `lan_ingress` `TC_ACT_SHOT`s that group's new flows — observed as domestic outages flapping dead→`revived via traffic` every few minutes. (Review note: the reserved slot 0/1 itself is structurally unwritable by the publisher — `checked_sub(UserBase)` — so the literal "outbound 0 pushed dead" path is not reachable in current code; the reachable harms are group ejection and group-slot death.)

`block` got the same exemption in #128 ("a blocked flow is policy working, not a health signal"); `direct` was left exposed.

## Why verdict immunity, not just probe changes

The kill path was **traffic-reported** failures, so probe-side exemptions alone cannot fix it. `direct` has no alternative to fail over to: marking it dead can only break traffic, never reroute it. This mirrors the existing `report_dial_latency` exemption ("local-egress latency is not node quality").

## Changes

- `mark_unavailable_internal`: early return for `DIRECT_NODE_ID | BLOCK_NODE_ID` — probe, traffic, and forced death paths all converge here, so the builtins can never be marked dead, never publish a dead state toward `OUTBOUND_CONNECTIVITY_MAP`, and are never health-filtered out of a group. direct's TCP probe (raw connect to the bootstrap target) still feeds latency ranking; only the liveness verdict is suppressed. UDP probes were already exempt.
- `record_dial_failure`: extend the `BLOCK_NODE_ID` exemption to `DIRECT_NODE_ID` so swarm noise cannot strike-demote `direct` in URLTest groups either.
- Custom-URL groups: a member whose current leaf is a builtin now advances the `(tag, url)` liveness machine instead of returning a state-neutral success — a tag that died under a previous non-builtin leaf recovers instead of filtering the member forever.
- Tests: `test_direct_ignores_failure_marks` drives every failure API across the full domain×family product past the registration grace (verified revert-sensitive); the block test now calls `mark_dead`; `test_builtin_leaf_probe_recovers_url_state` pins the custom-URL recovery incl. its two-success hysteresis and no fake latency sample.
- Docs: fail-closed sections (AGENTS.md + design/reference, en+zh) now state that a group containing a `direct`/`block` builtin never fails closed, since the builtins are never marked dead.

Considered and skipped: a kernel-side `OUTBOUND_DIRECT` exemption in `wan_outbound_is_alive` (belt-and-braces, but requires an eBPF object rebuild; the userspace choke point makes a dead publish unreachable).

## Verification

- Workspace suite green (honk-outbound alive 59/59 incl. the new tests); clippy `--all-targets -D warnings` clean.
- Deployed to the incident gateway: before the fix, 20 death/revival flaps in 25 min under BT load; after, `direct` stays alive (zero revivals across every 12-min observation window, ~800 direct dial failures each).


Glassyiris (commented):
reviewed

PR #133, opened 2026-09-04
fix(dns): gate proxied upstream dials by the carrier's health domain

Glassyiris:
## Problem

A node whose airport blocks the UDP `:53` check target (`udp_check_dns`) gets `DnsUdp` marked dead while its real data path (TCP and QUIC 443) is fine. Proxied DNS upstream dials were health-gated on the *protocol's nominal* domain instead of the carrier the dial actually uses:

- `quic://` / `h3://` were gated on `DnsUdp` — wrong, they run QUIC over the leaf's `PacketTransport`, which the `DataUdp` domain measures. Now gated on `DataUdp`.
- `udp://` (incl. `tcp+udp://`) was gated on `DnsUdp` — wrong in the other direction: proxied UDP is deliberately carried as pooled TCP-DNS (`transports.rs:90`, `query_udp_via_proxy`), so TCP health and the TCP score bucket apply. Now gated on `Tcp`.

Both mis-gates permanently exclude a working node ("no available node") and, in the old `udp://` case, also trained the UDP score bucket with physically-TCP exchanges. Direct dials never reach this context (no health gate).

Follows the split-domain model from #130 and the documented "feedback follows the carrier actually attempted" rule. External review caught the `udp://` half (proxied carrier is TCP-DNS) — credit to that reviewer; verified line-by-line before adopting.

## Changes

- `upstream_pool::routing::target_context`: `Udp → (Tcp, Tcp)`, `Quic | H3 → (Udp, DataUdp)`, `Tcp | Tls | Https → (Tcp, Tcp)` unchanged, with a why-comment.
- Tests: `implicit_quic_routes_use_data_udp_health` (DnsUdp dead → DoQ/DoH3 still select, DataUdp context) and `implicit_udp_route_follows_tcp_carrier_health` (DnsUdp dead → proxied `udp://` still selects, Tcp context).
- Docs: `doc/{en,zh}/design/groups.md` carrier/bucket sentence now states proxied UDP is TCP-carried.

## Verification

- `cargo test -p honk-core --lib` 943/943; honk-outbound 635/635; clippy `--all-targets -D warnings` clean.
- Live smoke on the incident build: `h3://dns.google` through an hy2 node answers correctly.

(The brotli 9 bump was split out of this PR — unrelated to the DNS fix; it lives on branch `deps/brotli-9` for a separate dependency decision.)


Glassyiris:
reviewed

PR #134, opened 2026-09-04
fix: reject three inputs that fail open or panic

Zakkaus:
* A `check_interval` that fails to parse becomes zero, and a zero period panics
  `tokio::time::interval`, so the periodic health-check loop dies at startup and outbound state
  then changes only from real traffic. `reload/warm.rs` already floors the same value at 10s; the
  probe path does not. `ControlPlane::reload_runtime_config` has no validating caller, so it now
  validates its candidate there too. The check stays on that public entry rather than inside the
  shared reload transaction, because subscription merges reach the transaction directly and
  legitimately publish nodes a whole-config check would reject.
* `security=reality` without `pbk` leaves `reality_public_key` as `None`, and
  `parse_reality_config` reads that as "no REALITY", so the node dials ordinary TLS with the
  configured SNI. The invariant is enforced in `VlessConfig::validate`, so it covers structured
  TOML/YAML/JSON nodes as well as share links, and it keys on whether a REALITY field is present
  rather than non-empty — an empty `sid` is documented as valid and would otherwise slip through.
* `ProtoDecoder` guards a length-delimited field with `self.pos + len > self.data.len()`. The
  length is asset-supplied and the release profile has no overflow checks, so a ten-byte varint
  encoding `u64::MAX` wraps past the guard and panics on the slice. `read_varint` also shifted the
  tenth byte before its overflow check, so a value wider than `u64` arrived as a small length; it
  now rejects that byte instead.

Each commit carries its test and, where the behaviour is documented, both language references.

`cargo fmt --all --check` and `cargo clippy --workspace --all-targets` are clean;
`cargo test --workspace --no-fail-fast -- --skip test_routing_with_config_dae` is 941 passed,
2 failed. The two failures are `routing::tests::test_geosite_cn_route` and `test_geosite_route`,
which need a `geosite.dat` that this environment does not provide; they fail identically on `main`
here. `cargo bench -p honk-core --bench reload --no-default-features --features reload-alloc-bench`
reports `allocations=28 bytes_allocated=68285 flag_writes=0` against `18` and `683` on `main`:
the validation on the public reload entry costs about 67 KiB per call, bounded by the size of the
configuration being validated.

### DIFF
--- a/crates/honk-core/src/control/reload/transaction.rs
+++ b/crates/honk-core/src/control/reload/transaction.rs
@@ -128,8 +128,16 @@ impl ControlPlane {
     }
 
     /// Publish an explicit runtime configuration through the same transaction
-    /// used by SIGHUP and subscription refreshes.
+    /// used by SIGHUP and subscription refreshes. This entry point has no caller that has
+    /// already validated, so it validates here, the way the SIGHUP handler does. The check
+    /// stays on this public entry rather than inside the shared transaction: subscription
+    /// merges reach that transaction directly and legitimately publish nodes a whole-config
+    /// check would reject.
     pub async fn reload_runtime_config(&self, new_config: Config) -> bool {
+        if let Err(error) = new_config.validate() {
+            error!(%error, "runtime reload rejected: invalid configuration");
+            return false;
+        }
         let drain = Arc::clone(&self.drain_tracker);
         self.apply_runtime_config(new_config, &drain).await
     }
--- a/crates/honk-config/src/config.rs
+++ b/crates/honk-config/src/config.rs
@@ -598,6 +598,14 @@ impl Config {
             .client_subnet_mode()
             .map_err(|error| crate::ConfigError::Validation(error.to_string()))?;
 
+        // A duration that fails to parse becomes zero, and a zero period makes
+        // tokio::time::interval panic, taking the health-check loop down at startup.
+        if self.global.check_interval_secs == 0 {
+            return Err(crate::ConfigError::Validation(
+                "global.check_interval must be a positive duration, such as 30s".into(),
+            ));
+        }
+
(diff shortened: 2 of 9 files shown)


crates/honk-core/src/control/reload/transaction.rs
```diff
@@ -128,8 +128,13 @@ impl ControlPlane {
     }
 
     /// Publish an explicit runtime configuration through the same transaction
-    /// used by SIGHUP and subscription refreshes.
+    /// used by SIGHUP and subscription refreshes. The candidate is validated
+    /// here because this entry point has no caller that already did so.
     pub async fn reload_runtime_config(&self, new_config: Config) -> bool {
+        if let Err(error) = new_config.validate() {
```
Glassyiris, line 137: 将  new_config.validate() 引入reload后增加 reload 中的内存，但是是有界的，预期也不会超过 1Mib 运行跳过 CI

Glassyiris (commented):
没问题！

crates/honk-core/src/control/reload/transaction.rs
Zakkaus: Measured with `cargo bench -p honk-core --bench reload --no-default-features --features reload-alloc-bench`: `allocations=28 bytes_allocated=68285` against `18` and `683` on `main` — about 67 KiB per call, bounded by the size of the configuration being validated rather than growing with reload count, which matches your read.

One thing worth recording since the diff moved: the check is deliberately on this public entry rather than inside the shared reload transaction. I tried it in the transaction first, after the effective-unchanged early return so the benchmark path stayed free, and `public_reload_surface_publishes_a_coherent_runtime` failed — subscription merges reach that transaction directly and legitimately publish nodes a whole-config check rejects. Keeping it here leaves those paths untouched.

PR #136, opened 2026-09-05
fix(core): unlink only the BPF pins this instance created

Zakkaus:
`remove_nonpersistent_pins` enumerated the whole pin root and deleted every entry except
`UDP_DECISION_SEQUENCE`, with `remove_dir_all` for directories. The root defaults to
`/sys/fs/bpf`, which every other BPF consumer on the host also pins into, so an orderly honk
shutdown — and `Drop` — removed pins belonging to Docker, CNI, systemd or anything else there.

The backend now records the map names it pinned during attach and unlinks exactly those. Crash
recovery is unaffected: attach already removes a stale pin by name before re-pinning each map, and
removes `LISTEN_SOCKET_MAP` explicitly, so a successor reclaims its own names without the sweep.
`UDP_DECISION_SEQUENCE` is pinned through `map_pin_path` and skipped by that loop, so it is absent
from the recorded list and survives by construction rather than by a name comparison in cleanup.

One trade-off worth stating: a pin whose name the current object no longer contains — renamed,
removed, or behind a disabled feature — is no longer swept away and lingers until something removes
it. The sweep used to catch that. Names are recorded when the attach loop claims them rather than
after a successful pin, so a stale pin whose unlink and re-pin both fail is still cleaned up later;
what is lost is only the names this build does not know about. Leaving a stale pin of our own is
visible and recoverable; deleting another program's pin is not.

Three adjacent hazards are pre-existing and deliberately untouched here: attach still unlinks a
foreign pin that collides with one of our map names, the `Justfile` recipes still operate on the
flat root, and `eject` unlinks without detaching hooks and currently has no caller.

This does not change the pin layout, the `--bpf-pin-root` contract, or the documented cleanup
behaviour — `AGENTS.md` already describes cleanup as removing ephemeral pins while preserving the
rollback-compatible allocator, which is what this now does precisely. An owned `<root>/honk`
subtree would be the stronger boundary, but it changes the CLI meaning and the allocator's
rollback path, so it is left as a separate decision.

Verified with `cargo fmt --all --check`, `cargo clippy --workspace --all-targets` (clean), and
`cargo test --workspace --no-fail-fast -- --skip test_routing_with_config_dae` at 942 passed,
2 failed — the two geosite tests that need an asset this environment does not provide, failing
identically on `main`. The behaviour itself is proved by a root-gated test rather than the mock,
which does not touch bpffs: `cleanup_leaves_foreign_pins_under_a_shared_pin_root` pins a foreign
map in a shared root, records its map id, runs load and cleanup, and asserts the foreign id and the
allocator pin are still there. Run under `--features ebpf --ignored`, it passes alongside the other
root-only eBPF tests; `control::tests::nfqueue_tc_netns_direct_proxy_contract` fails here with a
bare `No such file or directory` on an unmodified tree as well.

### DIFF
--- a/crates/honk-core/src/ebpf/real/attach.rs
+++ b/crates/honk-core/src/ebpf/real/attach.rs
@@ -123,6 +123,7 @@ impl RealEbpfBackend {
         }
         let mut bpf = loader.load(obj)?;
         syscall::validate_loaded_udp_decision_sequence(&bpf)?;
+        let mut pinned_maps = Vec::new();
         for (name, map) in bpf.maps() {
             // aya exposes ELF internal sections (.rodata, .bss, etc.) as maps.
             // These cannot be pinned to bpffs; skip them to avoid noisy warnings.
@@ -134,6 +135,9 @@ impl RealEbpfBackend {
                 continue;
             }
             let pin_path = pin_root.join(name);
+            // Recorded before the attempt, not after it: a stale pin whose unlink and re-pin both
+            // fail still belongs to us, and cleanup is the next chance to remove it.
+            pinned_maps.push(name.to_owned());
             if let Err(error) = std::fs::remove_file(&pin_path)
                 && error.kind() != std::io::ErrorKind::NotFound
             {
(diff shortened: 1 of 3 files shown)


Glassyiris (commented):
reviewed

Zakkaus:
> reviewed

還不睡覺

Glassyiris:
tested on vyos #136 

可能会有的 remove→pin 竞争 未在多次restart中复现

PR #137, opened 2026-09-05
fix(score): preserve feedback across warm dials and UDP retirement

Glassyiris:
## Summary

- Make ordinary and post-race TCP ready/bare pool refills streak-neutral: keep setup-quality evidence without clearing or increasing real-flow failure streaks and exploration backoff.
- Start admission-scoped TCP Score feedback before logical opens on reused sessions and QUIC connections, rather than after completion. Preserve the one-shot callback, cold physical-admission boundary, and completed-path fallback; invoke the QUIC hook outside the cache lock.
- Capture UDP terminal outcomes before the driver's synchronous health-death callback can retire its endpoint. Preserve failures from first sends, retained followers, and established traffic while retaining the original `io::Error` and existing cancellation, shutdown, congestion, idle, and QUIC-stall precedence.
- Update the bilingual lifecycle documentation and module map.

## Regression evidence

Against the unchanged production baseline `b98c8059ced6ea727af584ebb959df3816137c2c`, test-only overlays fail at both warm logical-start assertions and at the real UDP driver's post-retirement Score selection. The corrected implementation passes those reproductions.

The final UDP regression exercises the real pool driver, the 50th DataUdp traffic-failure callback, endpoint retirement, Score cleanup, and subsequent node selection at all three capture boundaries: first send, retained follower, and steady send. The existing neutrality test now checks immediately after neutral success so a subsequent failure cannot conceal an incorrect state transition.

## Local validation

- `cargo fmt --all`
- Focused warm logical-start tests: **2 passed**.
- Focused core control, Score/Clash, and NFQUEUE suites passed.
- Strengthened exploration-backoff test and consolidated three-stage UDP regression passed.
- `sh ci/outbound-ci.sh`: **ALL GREEN**.
- `cargo test --workspace --no-fail-fast -- --skip test_routing_with_config_dae`: **1942 passed, 4 ignored, 1 filtered**. The named routing exclusion is the existing repository gate, not a new exclusion.
- `cargo clippy --workspace --all-targets -- -D warnings`: passed.

## Scope and evidence limits

No Score formula, exploration parameter, configuration schema, physical-dial cap, kernel/NFQUEUE ABI, packet framing, or replay behavior changes. Existing eager UDP/probe/refill timing is not redesigned.

The four refill branches are verified through production code tracing and shared-model tests, not a live refill integration scenario. The new warm tests prove callback timing; exactly-once behavior retains the existing shared `FnOnce`/`take` mechanism. The UDP regression directly exercises steady send; receive converges on the same post-select capture point. No real eBPF/netns, remote-upstream interoperability, or hardware-performance claim is made.


Glassyiris (commented):
reviewed by @Glassyiris 

code: @Glassyiris 
tests/ci/pr: GPT 6 astra

PR #138, opened 2026-09-05
fix(core): DNS SERVFAIL observability + mac-bypass doc note

Glassyiris:
Follow-ups from the production DNS outage investigation (58-min upstream-transport wedge, reload-resistant, cleared by restart).

## Changes

- `fix(core)`: the SERVFAIL path logged at `debug` and discarded the error — a total upstream outage was invisible at the default level. Now `warn!` with `%error`, rate-limited to one line per 10s on a **monotonic** clock (wall-clock steps must not mute it).
- `docs(routing)`: a `mac(...)` bypass does not exempt a client from port-53 DNS interception — the LAN fast path runs before the routing engine, so no routing rule can exempt DNS; only a specifically bound non-honk local `:53` listener takes precedence over the fast path.

## Dropped after external review (credit: reviewer)

- The `Notify::enable()` hardening and the DoQ/DoH3 constructor timeout were withdrawn: under pinned tokio 1.53.1, `notify_waiters()` cannot be lost to a created-but-unpolled `Notified` (counter snapshot), and the timeout cannot fire on today's fully synchronous constructors — both were behavior-preserving at best and read as incident fixes they were not.

The wedge's originating mechanism remains open (kernel/attach state born during a WAN renumbering at process start is the leading candidate; userspace code paths were exonerated by review). Next occurrence: capture `tc filter show` + conntrack for a wedged flow before restarting (the on-box `diag.sh` collector now records `dns_upstream_fail` counts).


Glassyiris:
想尝试完全使用AI 进行修复，但是K3 表现不佳

PR #141, opened 2026-09-05
fix(paths): use FHS data directory and discover honk geo assets

### LINKED ISSUE #139: paths: the default data_dir /var/share/honk is not an FHS location
## What

`crates/honk-config/src/paths.rs:11` defaults the runtime data directory to a
path outside the FHS hierarchy:

```rust
pub const DEFAULT_DATA_DIR: &str = "/var/share/honk";
```

FHS 3.0 lists what may appear under `/var`: `cache`, `lib`, `local`, `lock`,
`log`, `opt`, `run`, `spool`, `tmp`, plus the optional `account`, `crash`,
`games`, `mail`, `yp`. `share` is not among them, and the string `/var/share`
does not occur anywhere in the standard
(<https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html>). `/usr/share` is
the architecture-independent share hierarchy; `/var` has no counterpart.

## Why it is not only a packaging concern

`global.data_dir` is configurable and honk creates the directory at startup, so
nothing is blocked. But the quickstart tells every reader to create it by hand
and to run the service out of it:

```
doc/en/how-to-start.md:179  sudo install -d -m 0700 /etc/honk /var/share/honk
doc/en/how-to-start.md:221  sudo curl -fL --retry 3 -o /var/share/honk/geosite.dat \
doc/en/how-to-start.md:240  WorkingDirectory=/var/share/honk
```

So anyone following the documented install ends up with a `/var/share` on their
system, not just distribution packagers. Downstream packaging is where it
surfaces first, because the installed file list is checked against the FHS.

## Scope of the change

Smaller than it looks. Outside tests the path is written in four places:

```
crates/honk-config/src/paths.rs:11   pub const DEFAULT_DATA_DIR
crates/honk-tool/src/geo.rs:19       doc comment
crates/honk-tool/src/geo.rs:46       doc comment
crates/honk-tool/src/geo.rs:114      user-facing "not found" message
```

(`paths.rs:125`, `paths.rs:129` and `cachedb.rs:799` also match, but all three
are inside `#[cfg(test)]` modules starting at `paths.rs:117` and
`cachedb.rs:781`.)

Documentation carries 24 lines across 10 files: `doc/{en,zh}/` ×
`how-to-start.md` (5 each), `reference/global.md` (3), `configuration.md` (2),
`reference/cli.md` (1), `reference/subscription.md` (1).

## Suggested change

Default to `/var/lib/honk` and treat `/var/share/honk` as a legacy location.

A helper with the right semantics already exists —
`resolve_artifact_path_with_legacy()` (`crates/honk-config/src/paths.rs:67`),
documented for "retaining an existing legacy location during upgrades" — but it
is not wired for this. Its only two callers pass a different legacy concept:

```
crates/honk-core/src/cachedb.rs:770      legacy = cache.db beside the legacy config dir
crates/honk-core/src/subscription.rs:57  legacy = an existing ./.sub
```

Neither passes anything under `/var/share/honk`, so moving `DEFAULT_DATA_DIR`
on its own would strand an installed `cache.db`. Making the move safe means
passing the old `/var/share/honk/<name>` as an additional legacy candidate at
those two sites.

One class of artifact has no legacy path at all. `find_dat()` resolves
`geosite.dat` and `geoip.dat` through plain `resolve_artifact_path()`
(`crates/honk-core/src/routing/geo.rs:413`), not the legacy-aware helper — and
those are the two files the quickstart tells the reader to download into the
data directory (`doc/en/how-to-start.md:221`, `:223`). A default change moves
the directory out from under them with nothing to fall back to, so they need
either the same legacy treatment or an explicit line in the release notes.

## Where this came from

Packaging honk for the gentoo-zh overlay. The ebuild has to create
`/var/share/honk` itself, because otherwise:

```
WARN Runtime data directory is unusable; using process working directory
     requested=/var/share/honk error=Permission denied (os error 13)
```

and the fallback working directory is `/` under both the OpenRC script and the
systemd unit.

### LINKED ISSUE #140: geo: the asset search covers dae's share directories but not honk's
## What

`crates/honk-core/src/routing/geo.rs:406-424` resolves `geoip.dat` and
`geosite.dat` in this order:

1. `$DAE_LOCATION_ASSET/<name>`
2. `resolve_artifact_path(<name>)`, i.e. `<data_dir>/<name>`
3. `./<name>`
4. `/usr/local/share/dae/<name>`
5. `/usr/share/dae/<name>`
6. `/etc/dae/<name>`

Steps 4 to 6 are dae's asset locations. honk's own `/usr/share/honk` and
`/usr/local/share/honk` are not searched. This is the engine's own path, not
only the tool's: `routing/geo.rs:101-102` calls `find_geosite_dat()` and
`find_geoip_dat()`, both thin wrappers over `find_dat()`.

## Why it matters

`/usr/share/<package>` is where a distribution puts read-only shared assets, so
it is the first place a packager reaches for. The first version of our ebuild
symlinked the v2ray geo data into `/usr/share/honk/` and it was never found.

Nothing in the log says so, because `capture_source()` treats the failure modes
asymmetrically:

```
routing/geo.rs:151  find() returned None -> GeoSource::Missing, no log
routing/geo.rs:156  read failed          -> tracing::warn!
routing/geo.rs:393  geosite parse failed -> tracing::warn!
routing/geo.rs:569  geoip parse failed   -> tracing::warn!
```

A `Missing` source only reaches `bytes() -> None` (`:84`) and the fingerprint
(`:132`); it never surfaces to the operator. So an asset the search cannot
locate is indistinguishable at runtime from a geosite or geoip rule that matches
nothing, while an unreadable or corrupt file at the very same path is reported.

## The sibling package shows the shape

`net-proxy/dae` in the same overlay symlinks its geo data into the directory dae
searches:

```bash
dosym -r /usr/share/v2ray/geoip.dat /usr/share/dae/geoip.dat
dosym -r /usr/share/v2ray/geosite.dat /usr/share/dae/geosite.dat
```

That works because `/usr/share/dae` is on dae's own search list. honk inherited
the list, dae's three entries included, without adding its own — so the same
packaging shape, applied to honk, lands the files where nothing looks.

## Suggested change

Add honk's own directories ahead of dae's:

```rust
for directory in [
    ".",
    "/usr/local/share/honk",
    "/usr/share/honk",
    "/usr/local/share/dae",
    "/usr/share/dae",
    "/etc/dae",
] {
```

Putting them before dae's three is not a regression. No released version has
ever looked in `/usr/share/honk`, so the only files that can appear there are
the ones a packager deliberately installed for honk.

## A second site has to change with it

`crates/honk-tool/src/geo.rs:113-115` repeats the whole list in its error
message:

```rust
"{name} not found (tried $DAE_LOCATION_ASSET, /var/share/honk/{name}, ./{name}, \
 /usr/local/share/dae, /usr/share/dae, /etc/dae) — pass --file PATH"
```

Changing only `routing/geo.rs` leaves this message describing a search order the
code no longer performs. The doc comments at `crates/honk-tool/src/geo.rs:19`
and `:46` spell out the same list.

Glassyiris:
## Summary

Closes #139.
Closes #140.

- Default `global.data_dir` to `/var/lib/honk` for new installations.
- Preserve existing writable artifacts and read-only dependencies through the shared legacy resolver: configured data-directory copy → `/var/share/honk` copy → caller-specific legacy candidate → configured directory for new artifacts. Absolute child paths stay explicit; custom data directories follow the same fallback order. Existing cache databases and subscription stores continue updating in place; there is no automatic migration. Relative logs remain rooted in the configured directory.
- Geo lookup uses the first regular file in this order: `$DAE_LOCATION_ASSET`, configured data directory, `/var/share/honk`, working directory, `/usr/local/share/honk`, `/usr/share/honk`, `/usr/local/share/dae`, `/usr/share/dae`, `/etc/dae`.
- Warn with the filename when a referenced geo asset is missing; leave unused assets silent.
- Synchronize CLI help/full-path errors, example config, bilingual installation/path references, and the AGENTS module map.

## Verification

- Actual API/CLI upgrade smoke in a private mount namespace: default and custom data roots × preferred/old-root/caller-legacy/fresh scenarios. Verified retained SQLite selector reads and writes, subscription-store reuse, read-only dependency contents, creation-only log paths, and absolute path precedence.
- Actual `honk-tool` smoke for geosite and geoip: all nine lookup priorities with relative package-style symlinks, missing-file diagnostics, and help output.
- `cargo fmt --all`
- `env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY sh ci/dns-ci.sh` — ALL GREEN (config, DNS, DNS runtime integration, control, outbound tests and Clippy).
- `cargo clippy -p honk-tool --all-targets -- -D warnings`
- Focused core tests: cachedb 18 passed; subscription 32 passed; geo scan 12 passed.
- `cargo test -p honk-tool` — 13 passed.

Temporary upgrade driver/scripts/fixtures were removed after verification. Real eBPF datapath deployment and cross-architecture release builds were not run for this userspace path-resolution change.


Glassyiris (commented):
reviewed

Zakkaus:
Two files still describe the pre-change search order, and neither is in this PR.

`doc/en/configuration.md:45`:

> Geo lookup checks an existing `$DAE_LOCATION_ASSET/<file>` first, followed by the effective data directory, working directory, and standard dae asset directories.

`doc/zh/configuration.md:45` carries the same sentence ("dae 标准资源目录"). After this change the order is `$DAE_LOCATION_ASSET` → data directory → working directory → honk's share directories → dae's legacy directories.

They are easy to miss: all eight files this PR touches contain a literal `share/dae`, while these two describe the order in prose with no path in it, so a sweep for that string does not reach them. `git grep -n -i 'dae asset\|asset director\|资源目录' doc` finds them.

Nit, take or leave: `capture_source` taking a `&str` means each asset filename is now written in two places — `geo.rs:101` and `geo.rs:407` for geosite, `geo.rs:102` and `geo.rs:589` for geoip. Passing the finder kept the name inside `find_geosite_dat` only.

The rest looks right to me. I checked that the new warning cannot flood: every production `GeoSourceSet::load` caller is a constructor or reload — `lib.rs:1055`, `build_forwarder_from_config` reached once from `resolver.rs:26`, `reload/transaction.rs:208`, and the `From`/`Router::new` pair in `routing/mod.rs`, with the `dns/routing.rs` calls under `#[cfg(test)]`. Both honk-tool sites were updated, so the "not found" message no longer names an order the code does not perform.

Glassyiris:
保留当前， 在正式版前会去掉对dae 目录和环境变量的支持

Glassyiris (commented):
reviewed

code/test　@Glassyiris 
ci/doc: gpt-6-astra max

Zakkaus:
The paths commit removes the three assertions that pinned the default data directory, rather than updating their expected string:

- `crates/honk-config/src/paths.rs` — `artifacts_root_relative_paths_in_the_data_directory`
- `crates/honk-core/src/cachedb.rs` — `relative_cache_path_uses_the_data_directory`
- `crates/honk-config/src/parser/tests.rs` — `assert_eq!(default.global.data_dir, "/var/share/honk")`

On this head, `/var/lib/honk` appears in four places: the constant itself and three doc/error strings in `honk-tool`. No test reads it. The remaining `paths.rs` tests all call the `*_from` variants with an explicit directory, or pass absolute paths, and both shapes return before `data_dir()` is consulted — so editing `DEFAULT_DATA_DIR` now leaves the suite green.

If the reason for deleting rather than updating was the `OnceLock`, that does not apply here: `set_data_dir` has no caller in any test, only `lib.rs:544`, so those assertions were not order-dependent and only the expected string needed to change.

This seems worth one test given you plan another pass over this area before the stable tag to drop the dae directories and `DAE_LOCATION_ASSET`. Asserting the constants directly cannot race, because it never touches `data_dir()`:

```rust
#[test]
fn default_data_dir_is_the_fhs_location() {
    assert_eq!(DEFAULT_DATA_DIR, "/var/lib/honk");
    assert_eq!(LEGACY_DATA_DIR, "/var/share/honk");
}
```

The rest of the commit checks out from here. The three call sites still on the non-legacy `resolve_artifact_path` are all correct — `lib.rs:507` is the log file, a write target; `geo.rs:417` is covered by `LEGACY_DATA_DIR` being first in the `find_dat` loop; `subscription.rs:74` only computes `preferred` for the comparison behind the warning, while resolution goes through `default_store_root`. `how-to-start.md` is fully migrated including `install -d`, both `curl -o` targets and `WorkingDirectory`, and no document still calls `/var/share/honk` the default.

Glassyiris:
all clear, merged

PR #143, opened 2026-09-06
fix(core): close reviewed datapath and runtime trust gaps

Glassyiris:
## Problem

The [external review](https://paste.gentoozh.org/md/sWgz-+KpyVb.md) exposed routing and NFQUEUE failure-boundary bugs, DNS validation/cache errors, stale health state, and TLS/protocol correctness issues. Some claims described existing limitations rather than bugs.

## How I fixed it

- Separate packet-local NFQUEUE failures from fatal metadata/verdict errors; enforce fragment handling and positive host-local address authority.
- Reuse one locality snapshot per reconciliation and insert only missing map keys, while retaining failure revocation and repair.
- Validate DNS transaction IDs and predicates, clamp UDP responses, and fix TTL invalidation and domain projection.
- Bind health/probe state to the current generation, update every affected group, and repair startup scheduling and unchanged-config reloads.
- Reject unsupported configuration, isolate QUIC tickets by endpoint and verification policy, and distinguish valid VLESS EOF from truncation.
- Remove inert AnyTLS limits, obsolete health mappings, and redundant tests without weakening resource bounds. Update bilingual documentation.

## Verified

Head: `8dcc09b522deb50ca67e3fe976868f8d33f67c99`.

- Workspace: **1,970 passed**, 4 ignored; existing `test_routing_with_config_dae` exclusion retained.
- eBPF-feature core: **1,017 passed**, 9 ignored. eBPF-enabled all-targets Clippy and workspace format checks passed.
- Locality follow-up on the x86 host: **6 root lifecycle tests passed**; a kernel smoke confirmed missing/stale-key repair with an unchanged snapshot and no writes to an unchanged frozen map.
- At `25e6b32`: DNS/health process smokes, **22 eBPF program loads**, **35 TC checks**, and **8 root-gate tests** passed on x86 Linux 7.1.8.
- Five-lane adversarial and code-quality reviews at `25e6b32` found no remaining verified P1/P2; the locality follow-up preserves the reviewed failure boundaries.

GitHub CI is still running. ARM/OpenWrt, arbitrary VLAN trunks, and full production deployment are not claimed as verified.


Glassyiris:
截止 [25e6b32](https://github.com/daeuniverse/honk/pull/143/commits/25e6b323f704a70d4b73d963519708725343e107) 都为 Gpt-6-astra， 当前在进行人工重审， 之后commit 仅会由 @Glassyiris

Zakkaus:
Reviewed `becdbd30..25e6b323` across five lanes, then re-checked both findings against the current head `8dcc09b`. Both still apply there; line numbers below are on `8dcc09b`.

## CONFIRMED — the IPv6 host fix drops a separate-port override

`crates/honk-config/src/node.rs:150`:

```rust
if let Some((host, port)) = self.address.rsplit_once(':')
    && port.parse::<u16>().ok() == Some(self.port)
{
    return host;
}
&self.address
```

The base was unconditional — `self.address.split(':').next()` — which is what broke bare IPv6, so handling `[...]` and a parsed `Ipv6Addr` first is right. The equality on the IPv4/hostname branch is the part that changes behaviour beyond that fix.

Trigger: a structured node with an omitted or empty `host`, an `address` carrying a port, and a different effective `port`:

```json
{"name":"local-socks","protocol":"socks5","address":"127.0.0.1:1080","port":2080}
```

Base produced host `127.0.0.1` and dialled `127.0.0.1:2080`. The head returns `127.0.0.1:1080` as the host, so a consumer building `format!("{}:{}", node.host(), node.port)` gets `127.0.0.1:1080:2080`. `Socks5Handler::dial` and `udp_association` both format it that way, and `util::connect_marked` strips only the last port, leaving `127.0.0.1:1080` for `bootstrap::resolve` — no IP-literal fast path, no system lookup, so the node fails before the SOCKS handshake. Neither `FlatNode::try_from` nor `Config::validate` enforces or normalises the new equality, so such a config still loads. Bracketed IPv6 keeps accepting the separate-port override; only this branch acquires the requirement.

The equality does not add protection: `port.parse::<u16>()` already requires the suffix to be numeric. Dropping `== Some(self.port)` and keeping the parse restores the previous separate-port precedence without touching the IPv6 work.

Established by copying the head's `Node::host()` body into a `rustdoc --test` harness and asserting the base endpoint, the new host, the malformed reconstruction, and the resolver failure, with matching-port, bare-address, bracketed-IPv6 and explicit-host controls passing. The production consequence is inference from reading those three functions; the engine was not launched.

## PLAUSIBLE — link-local locality key differs between the two sides

`crates/honk-core/src/ebpf/real/iface_watch.rs:124` publishes with the interface that owns the address:

```rust
let ifindex = if link_local {
    let Some(ifindex) = iface_address_ifindex(&ifname) else { continue };
    ifindex
} else {
    0
};
```

`crates/honk-ebpf/src/ingress.rs` keys the lookup on the receiving skb's ifindex instead. On a bridge these are not the same interface: if `br0` owns `fe80::1/64` and a slave receives the client traffic, the classifier looks up `(slave_ifindex, fe80::1)` while userspace published `(br0_ifindex, fe80::1)`. A wildcard `[::]:53` listener would then lose precedence and the port-53 branch would redirect the query into the transparent DNS path before it reaches the bridge hook. A bond has the same identity mismatch.

`8dcc09b` rewrote the snapshot and delta handling around this code but left the key derivation as above.

This one is reasoning from the two sides of the key, not an observed failure — we did not build a bridged topology, and you have the VM rig for it. If the slave hook is expected to see the bridge's ifindex here, then it is not a defect and I would rather know that than file it.

## Lanes that came back clean

Config rejections: each new rejection was checked for a legitimate previously-accepted reliance and none was found; the `group.final` deferral for subscription-provided nodes does not let the original direct fallback back in. Datapath: no confirmed fragment or NFQUEUE verdict-ownership defect. Runtime, protocol and sweep: no introduced regression established.

Two things reported earlier verify as fixed at this head: the shared leaf that only updated the first group's eBPF alive slot, and `full:` / `regex:` domain rules not reaching the eBPF projection. `65200bac` narrowing publication to a changed slice does not skip a slot that used to be rewritten wholesale.

PR #147, opened 2026-09-06
fix(subscription): refuse unsafe redirects and cap the fetched body

### LINKED ISSUE #144: subscription fetch follows redirects to plaintext and private addresses, and buffers an unbounded body
The subscription HTTP client is built with a timeout and a bootstrap resolver and nothing else, so a subscription host controls where honk connects, over which scheme, and how much honk buffers.

`crates/honk-core/src/subscription.rs:247`, on `becdbd3`:

```rust
let client = reqwest::Client::builder()
    .timeout(std::time::Duration::from_secs(30))
    .dns_resolver(std::sync::Arc::new(BootstrapDnsResolve))
    .build()?;
```

No redirect policy, no `https_only`, no destination filter, no body limit. `Cargo.lock` pins reqwest 0.13.4, whose defaults are `redirect::Policy::limited(10)` and `https_only: false`; its redirect gate permits both http and https, and rejecting a downgrade requires the flag that is not set. Config validation checks only the string as written — `crates/honk-config/src/config.rs:783` tests `starts_with("http://")` / `starts_with("https://")` — so it never sees the redirect target.

### Reproduced

Built each ref with `cargo build --locked -p honk-core` and ran the real `SubscriptionManager::fetch` inside a network namespace, with a temporary CA supplied through `SSL_CERT_FILE` so the local HTTPS origin authenticates. The client builder was not modified.

An HTTPS subscription whose endpoint answers `302 Location: http://10.203.0.1:PORT/sub` is followed over plaintext, and the share links served there become nodes: `nodes=1 first=plaintext-node`, with the plaintext receiver observed at `('10.203.0.1', 35365)`. The same holds for `http://127.0.0.1:PORT/sub`. So a subscription host — or anyone who can answer for it — can both strip TLS and reach addresses honk's operator did not publish it to.

### No size limit on the same path

`subscription.rs:277` buffers the whole body before parsing:

```rust
let content = response.text().await.map_err(reqwest::Error::without_url)?;
let nodes = parse_subscription_content(sub, &content)?;
```

`error_for_status()` checks status, not length, and reqwest 0.13.4's `text()` → `bytes()` → `BodyExt::collect` applies no `Limited` wrapper; the timeouts in `async_impl/body.rs` are time, not bytes. The parser then makes further whole-input allocations, and the on-disk restore path uses an uncapped `read_to_string`.

Measured: the production fetch accepted and parsed **33,554,473 bytes** on both refs, with `Content-Length` and again with chunked framing. That probe shows acceptance above 32 MiB; the absence of a limit is established by the aggregation code rather than by testing every size.

### Suggested shape

All four gaps are in one builder. A redirect policy that refuses a scheme downgrade and refuses non-public destinations unless the original URL was itself non-public, `https_only` when the configured URL is https, and a cap applied while reading rather than after, would close them together.

Verified against `becdbd3059ce` and also present on the head of #143 (`25e6b323f7`). Please say if any of this is intended for the alpha and I will stop reporting it.

Zakkaus:
## Summary

- Give the subscription client a redirect policy: at most 5 hops, no `https` to plaintext downgrade, and no redirect to a loopback, private, link-local, or unspecified literal address that the configured URL did not itself use. Each refusal returns an error rather than the redirect response, so the reason surfaces instead of a parse failure on a redirect body.
- Read the response with the cap applied while reading instead of buffering the whole body first, matching the external UI downloader's shape.
- Document both limits in the bilingual subscription reference.

Closes #144.

## Not covered

A redirect to a *hostname* that resolves to a private address is still followed; only a literal address stated in the redirect is refused. Filtering resolved addresses would belong in `BootstrapDnsResolve`, but that would also refuse a subscription self-hosted on a private address, which `Config::validate` accepts today. The comment on the predicate says so.

An initial `http://` subscription URL remains allowed, and a subscription whose configured URL is already a private literal may still redirect among private literals.

## Verification

- `cargo test -p honk-core --lib subscription::tests` — 33 passed, including a new address-predicate table and a cap-boundary test asserting that exactly `MAX_SUBSCRIPTION_BYTES` succeeds and one byte more fails.
- `cargo clippy -p honk-core --lib --all-targets` — clean.
- `cargo fmt --all --check` — clean.
- The two `routing::tests::test_geosite_*` failures in the full crate run are missing local geo assets and fail identically without this change.

`Policy::custom` is used rather than the `Policy::none()` plus manual loop in `clash_api/ui.rs`: that loop exists to re-route each hop and choose direct versus tunneled transport, which subscription fetches do not need.


Zakkaus:
Ran the redirect policy end to end, in the same shape that demonstrated the original behaviour: a throwaway CA, an HTTPS origin and a plaintext origin inside a network namespace, driving the real `SubscriptionManager::fetch` through a small example binary. `93.184.216.34/32` and `10.203.0.1/32` were added to `lo` so a public origin and a private destination both exist without leaving the namespace.

```
https downgrade to loopback   REFUSED  error following redirect | subscription redirected from https to plaintext
https downgrade to private    REFUSED  error following redirect | subscription redirected from https to plaintext
https to https, same host     OK       nodes=1 first=delivered
http to http, private origin  OK       nodes=1 first=delivered      (two hops)
public origin to private      REFUSED  error following redirect | subscription redirected to a private address
public origin to public       OK       nodes=1 first=delivered
32 MiB body                   REFUSED  subscription body exceeds 8388608 bytes
ordinary body                 OK       nodes=1 first=delivered
```

The four `OK` rows are the point of the run as much as the refusals: same-scheme redirects, multi-hop chains, and a subscription whose configured URL is itself a private literal all still work, so the policy does not over-refuse.

Two things this showed that the unit tests could not:

The private-address rule needed a public origin to be exercised at all. Both downgrade cases also target a private address, so they hit the scheme rule first and returned before the address check ran. With a public literal as the origin it fires on its own, which is the row above.

The refusal reason survives as a cause, not as the top-level message: reqwest reports `error following redirect` and the specific reason is one level down. It reaches an operator through the existing chain-walking error reporting, but a caller printing only `Display` would see the generic text.

The example binary used for this was not committed; the branch is unchanged.

Glassyiris (commented):
reviewed

PR #148, opened 2026-09-06
fix: close the two reload lock cycles

### LINKED ISSUE #146: two deadlocks reachable from reload: AliveDialerSet lock inversion, and a recursive read on the shared group manager
Two cycles, both reached from an ordinary reload racing something that runs concurrently with it. Line numbers are on `becdbd3`; both are also present on the head of #143 (`25e6b323f7`).

### (a) Reload versus the URLTest idle check — lock-order inversion

Three separate `parking_lot::RwLock`s inside one `AliveDialerSet`:

```
T = urltest_group_timeout
M = urltest_group_members
N = node_urltest_groups
```

Reload, through `sync_urltest_groups`, takes **T(write) → M(write) → N(write)** at `crates/honk-outbound/src/alive/mod.rs:1448`:

```rust
let mut timeouts = self.urltest_group_timeout.write();
let mut members_map = self.urltest_group_members.write();
let mut node_groups = self.node_urltest_groups.write();
```

The health scheduler, through `is_probe_suspended`, takes **N(read) → T(read)** at `alive/mod.rs:1533`:

```rust
let groups = self.node_urltest_groups.read();
match groups.get(&node_id) {
    Some(gs) if !gs.is_empty() => gs.iter().all(|g| self.is_urltest_group_idle(g)),
```

`is_urltest_group_idle` then reads T at `:1518`. The borrowed `gs` keeps N's guard alive across that call, so there is no drop boundary between the two acquisitions.

Cycle: health holds N(read); reload holds T(write) and M(write) and waits for N(write); health waits for T(read). Reload cannot release T until health releases N, and health cannot release N until reload releases T. M is held but is not needed to close it. The reload mutex serialises reloads against each other, not against the independently spawned health loop.

Trigger: a URLTest group with at least one node belonging only to URLTest groups (membership in any non-URLTest group excludes it from idle suspension), health checking running, and SIGHUP arriving while a check cycle visits that node on another worker. No node failure, no special server response, no custom check URL, and an unchanged effective configuration still reaches the table replacement. It is an interleaving, not every reload.

### (b) Selector interrupt callback versus reload's writer — recursive read

One lock: `SharedGroupManager = Arc<parking_lot::RwLock<Arc<GroupManager>>>` (`crates/honk-outbound/src/group/mod.rs:137`), acquired twice on one stack.

`crates/honk-core/src/clash_api.rs:626`:

```rust
s.group_manager
    .read()
    .set_selector_choice(&group_name, &body.name);
```

That temporary read guard lives until the statement ends, callbacks included. `group/state.rs:61` drops `selector_choice.write()` before `maybe_interrupt`, which does not release G, and `state.rs:136` invokes the installed callback. The callback's first line is `let gm = cell.read().clone();` on the same cell (`control/reload/connectivity.rs:173`) — the clone happens only after that second acquisition succeeds.

Reload ends its publication with G(write) at `control/reload/transaction.rs:483`.

Cycle: the API task holds G(read); reload requests G(write) and reserves writer admission while existing readers drain; the synchronous callback then requests G(read) again and queues behind that writer. The callback waits for reload, reload waits for the API task's outer guard, and that guard cannot drop until the callback returns. parking_lot documents this recursive-read hazard in `rwlock.rs:17-21`, and its writer reservation precedes reader draining.

Trigger: Clash API enabled, a Selector group with `interrupt_connections: true` and two valid members, an authenticated `PUT /proxies/<group>` changing the stored choice, arriving while SIGHUP reaches group-manager publication. No live connection is required — the callback re-reads the manager before it inspects tracked connections. The `drop(config)` at `clash_api.rs:619` is what lets reload reach G(write) during the update.

### Evidence

Reduced reproducers, not full-daemon SIGHUP races: the relevant methods were extracted separately from each ref with `parking_lot 0.12.5` as pinned, and scheduling barriers added after the first acquisition. The detector reported one two-thread cycle for (a) on both refs, with stacks in `sync_urltest_groups` and `is_probe_suspended → is_urltest_group_idle`, and one recursive self-cycle for (b) in `set_selector_choice → maybe_interrupt → callback`, with the writer also wedged.

Negative control for (b): cloning the manager and dropping the outer guard before calling the setter completed both the writer and the callback on both refs. Neighbouring snapshot-taking handlers already do exactly that; this setter call is the one that does not.

An earlier draft of this report cited the comment near `control/runtime.rs:181` — "Reload takes these locks in the same order" — as the evidence. That was wrong and is withdrawn: it concerns the config → eBPF ordering, which reload does preserve structurally at `reload/transaction.rs:479`. The inversion is the T/M/N one above, in a different structure.

Zakkaus:
## Summary

- `is_probe_suspended` copies the group names out of `node_urltest_groups` instead of holding that guard across `is_urltest_group_idle`, which takes `urltest_group_timeout`. Reload acquires the timeouts first and the node map last, so the two orders formed a cycle.
- `set_selector_choice` runs its persist, selector-change, and interrupt callbacks synchronously, and interrupt handling reacquires the shared group manager. The three callers that invoked the setter through a live read guard now clone the `Arc` first, as the neighbouring read handlers in `clash_api.rs` already do.

Closes #146.

## Sites changed

`clash_api.rs` `put_proxy` is the one the issue reported. `control/cache.rs` restores stored selector choices through the same shape and is fixed with it; leaving it would close the issue with the pattern still present.

## Observed but not changed

`control/connection/udp.rs:287` and `dns/upstream_pool/routing.rs:95` hold the manager guard across target selection, which can reach `maybe_interrupt` through URLTest and Fallback policies. Neither is an established deadlock on the ordinary SIGHUP path: UDP holds the config read guard while transactional reload takes the config write before the group write, and production DNS installs a pinned snapshot. They are reported here rather than changed, because the fix would be speculative surgery on a hot path. The guard audit is not clean; these two are what remains.

## Verification

Both tests were written to fail first. Each was run against the tree with the production fix reverted and again with it restored:

- `test_suspension_check_releases_node_groups_before_reading_timeouts` — without the fix, "the suspension check held node_urltest_groups while waiting for urltest_group_timeout" after the 5s deadline; with it, passes. It holds the timeouts, waits for the prober to enter, and then requires the node map to become free.
- `test_selector_switch_releases_the_group_manager_lock` — without the fix, "the handler held the group manager lock across the selector callbacks"; with it, passes. The probe runs inside a selector-change callback and reports whether a writer could acquire the lock.

An earlier version of the first test ran 200 concurrent iterations and passed against the unfixed code, so it was replaced: the race window is too small to hit by repetition, and a test that cannot go red is not a gate.

- `cargo clippy -p honk-core -p honk-outbound --all-targets` — clean.
- `cargo fmt --all --check` — clean.
- `cargo test --workspace` — the only failures are `routing::tests::test_geosite_*`, which need local geo assets and fail identically without this change.

No live reload or datapath run: both fixes are lock-lifetime changes and the tests assert that property directly.

### DIFF
--- a/crates/honk-core/src/clash_api.rs
+++ b/crates/honk-core/src/clash_api.rs
@@ -623,9 +623,10 @@ async fn put_proxy(
 
     // cache.db persistence runs through the group manager's persist
     // callback, wired by ControlPlane::init_cache_db.
-    s.group_manager
-        .read()
-        .set_selector_choice(&group_name, &body.name);
+    // The setter runs its callbacks synchronously and interrupt handling
+    // reacquires this lock, so the guard is released before the call.
+    let group_manager = s.group_manager.read().clone();
+    group_manager.set_selector_choice(&group_name, &body.name);
     StatusCode::NO_CONTENT.into_response()
 }
 
--- a/crates/honk-core/src/control/cache.rs
+++ b/crates/honk-core/src/control/cache.rs
@@ -19,14 +19,15 @@ impl ControlPlane {
         // callback so restoration does not rewrite the same values.
         {
             let config = self.config.read().await;
+            // The setter runs its callbacks synchronously and interrupt handling
+            // reacquires this lock, so the guard is released before the calls.
+            let group_manager = self.group_manager.read().clone();
             for group in &config.groups {
                 if group.policy == GroupPolicy::Selector
                     && let Some(node) = db.load_selector_choice(&group.name)
                 {
                     info!("cache.db: restored selector '{}' = '{}'", group.name, node);
-                    self.group_manager
-                        .read()
-                        .set_selector_choice(&group.name, &node);
+                    group_manager.set_selector_choice(&group.name, &node);
                 }
             }
         }
--- a/crates/honk-core/tests/clash_api_test.rs
+++ b/crates/honk-core/tests/clash_api_test.rs
@@ -375,6 +375,37 @@ routing {
 }
 
 #[tokio::test]
+async fn test_selector_switch_releases_the_group_manager_lock() {
+    // The callbacks must run without the shared-manager guard: interrupt
+    // handling reacquires it, which wedges behind a waiting reload writer.
+    let app = spawn_app("", "").await;
+    let observed = Arc::new(std::sync::atomic::AtomicBool::new(false));
+    {
+        let manager = app.state.group_manager.read().clone();
+        let cell = Arc::clone(&app.state.group_manager);
+        let observed = Arc::clone(&observed);
+        manager.set_selector_change_callback(Some(Arc::new(move || {
+            observed.store(
+                cell.try_write().is_some(),
+                std::sync::atomic::Ordering::SeqCst,
+            );
+        })));
+    }
+
+    let resp = http_client()
+        .put(app.url("/proxies/proxy"))
+        .json(&serde_json::json!({"name": "node-b"}))
+        .send()
+        .await
+        .unwrap();
+    assert_eq!(resp.status(), 204);
+    assert!(
+        observed.load(std::sync::atomic::Ordering::SeqCst),
+        "the handler held the group manager lock across the selector callbacks"
+    );
+}
+
+#[tokio::test]
 async fn test_proxies_structure_and_selector_switch() {
     let app = spawn_app("", "").await;
     let client = http_client();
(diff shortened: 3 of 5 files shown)


Glassyiris (commented):
reviewed @Glassyiris

Glassyiris:
tested on vyos, all clear

PR #152, opened 2026-09-06
fix(dns): finish terminal dispatch and reload error handling

Glassyiris:
## Problem

After #142, DNS errors could still fall through into generic forwarding, singleflight retained redundant completed state, and private-runtime preparation could panic during reload.

## How I fixed it

Made DNS dispatch terminal, consumed singleflight completion, and propagated runtime preparation errors. Added regressions and documented cancellation boundaries and the concise PR format.

## Verified

Before the code-identical rebase: 76 focused tests and the DNS gate passed. The musl/eBPF build passed VyOS UDP/TCP DNS, API, HTTPS, and SIGHUP smoke checks.


Glassyiris (commented):
test & verified by @Glassyiris 

only tests GPT-6-Astra

PR #154, opened 2026-09-06
fix(geo): warn when a geoip code expands to nothing

Zakkaus:
## Problem

`geoip_nets` says nothing when a code matches no network, while the geosite path in the same file warns for the same situation and its comment gives the reason — a code that expands to nothing silently disables its rule. The geoip `None` arm already warns when the index is unavailable, so the gap is narrow: an asset that loaded fine, and a code inside it that resolves to nothing. `dip(geoip:typo)` disables its rule and the log stays empty.

## How I fixed it

The expansion records the network count before the lookup and warns when nothing was added, matching the geosite wording and placement. It sits inside the shared expansion function rather than at a call site, so all four callers are covered at once.

## Verified

`geoip_code_expanding_to_nothing_warns_like_geosite` fails without the warning and passes with it, and asserts that a code which does match stays quiet so the warning cannot fire on the ordinary path. Only the diagnostic changes: a code that expands to nothing still contributes no networks, and whether that should instead fail the load is a separate question this does not answer. `cargo fmt --all -- --check`, `cargo clippy -p honk-core --all-targets`, and `cargo test -p honk-core --lib routing::geo` are clean.

### DIFF
--- a/crates/honk-core/src/routing/geo.rs
+++ b/crates/honk-core/src/routing/geo.rs
@@ -368,8 +368,17 @@ impl GeoAssets {
             }
             match &self.geoip {
                 Some(index) => {
+                    let before = nets.len();
                     if let Some(v) = index.get(&code.to_lowercase()) {
                         nets.extend(v.iter().cloned());
+                    }
+                    if nets.len() == before {
+                        // Same reason as the geosite path: a code that expands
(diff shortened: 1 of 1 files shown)


Glassyiris:
all clear.

PR #156, opened 2026-09-06
fix(socks5): bound the UDP associate negotiation

Zakkaus:
## Problem

`udp_associate` performs the same SOCKS5 greeting and RFC 1929 exchange as `handshake`, but without the five-second deadline that wraps it there, and its caller bounds only the connect that precedes it, so a server that accepts the control connection and then sends nothing parks the negotiation indefinitely.

## How I fixed it

The negotiation is wrapped in the same five-second budget; the address handling and relay-domain resolution that follow stay outside it, so a slow DNS answer for a domain-form relay is not folded into the handshake budget. The two negotiations remain duplicated — extracting one helper would change what `handshake`'s single budget covers, since it currently spans negotiation and CONNECT together, so that is worth proposing separately rather than folding in here.

## Verified

`udp_associate_gives_up_on_a_silent_peer` was written to fail first and does: without the deadline it reports that the call outlived its own timeout, and with it the negotiation deadline fires; it runs on a paused clock and costs no wall time. `cargo fmt --all -- --check`, `cargo clippy -p honk-outbound --all-targets`, and `cargo test -p honk-outbound --lib proxy::socks5` are clean.

### DIFF
--- a/crates/honk-outbound/src/proxy/socks5.rs
+++ b/crates/honk-outbound/src/proxy/socks5.rs
@@ -259,42 +259,51 @@ impl Socks5Handler {
         username: Option<&str>,
         password: Option<&str>,
     ) -> anyhow::Result<SocketAddr> {
-        let methods = if username.is_some() && password.is_some() {
-            vec![METHOD_NO_AUTH, METHOD_USERNAME_PASSWORD]
-        } else {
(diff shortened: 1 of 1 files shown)


Glassyiris (commented):
reviewed by @Glassyiris 

Okay'

Issue #157, opened 2026-09-06
an unparseable millisecond duration silently becomes 0


Zakkaus:
`parse_duration_ms` returns 0 for anything it cannot parse, so a duration a user believed they set becomes "no delay" with nothing in the log. Line numbers are on `ac4420f`.

`crates/honk-config/src/parser/mod.rs:1220`:

```rust
fn parse_duration_ms(s: &str) -> u64 {
    let s = s.trim();
    if s.ends_with("ms") { return s.trim_end_matches("ms").parse().unwrap_or(0); }
    if s.ends_with('s') {
        return s.trim_end_matches('s').parse::<f64>().map(|v| (v * 1000.0) as u64).unwrap_or(0);
    }
    s.parse::<f64>().map(|v| v as u64).unwrap_or(0)
}
```

Extracted and run against the inputs a user might write:

```
50ms  -> 50       0.5s  -> 500      50    -> 50
1m    -> 0        2h    -> 0        1min  -> 0
abc   -> 0        -5ms  -> 0
```

Its two callers are `global.check_tolerance` and `global.sniffing_timeout` (`parser/mod.rs:806,818`).

`m` and `h` are not an oversight in the grammar — `doc/en/configuration.md:25` states the split:

> Second-based durations accept bare seconds or `ms`, `s`, `m`, and `h` suffixes. Millisecond settings such as `check_tolerance` accept bare milliseconds, `ms`, or `s`.

So the documented behaviour is that `1m` is not accepted there. The reported problem is what "not accepted" does: `check_tolerance: 1m` loads without complaint as `0`, which means URLTest needs no latency improvement to switch members, so selection follows small ordering noise. The configuration reads as intended and the engine behaves as if the line were absent. `1min`, a plain typo, and a negative value do the same thing.

This is the shape #134 addressed for `check_interval`, where `.unwrap_or(0)` turned a unitless value into a zero interval. That one surfaced as a panic. This one does not surface at all.

Rejecting an unparseable duration would match how the second-based fields now behave. If a hard failure is too strong for these two fields, warning and falling back to the documented default would still leave the operator something to see.

Two things this issue is not asking for: `m` and `h` do not need to be accepted here — the split is documented and deliberate; and the two parsers do not need merging, since their rounding and precision rules differ for reasons.

Zakkaus:
[...]

PR #159, opened 2026-09-06
fix(config): reject millisecond durations the parser cannot read

Zakkaus:
## Problem

`parse_duration_ms` returned 0 for anything it could not read, so `check_tolerance: 1m`, `1min`, a typo, an empty value or a negative one loaded without complaint as zero — the configuration reads as intended while URLTest runs on its effective minimum of 1 ms (`group/policy.rs:220`, documented at `doc/en/reference/groups.md:42`) instead of the threshold the operator set. `inf` was worse: `as u64` saturates, so it became `u64::MAX`. Reported as #157.

## How I fixed it

The grammar moves to `types::parse_duration_ms`, beside `types::parse_duration_secs`, which already owns duration parsing and already returns `Option`; non-finite and negative values are filtered there before the cast. `parser/mod.rs` keeps only the wrapping into `ConfigError::Parse` naming the setting, which is the shape `parse_checked_bool` already has. The two callers, `global.check_tolerance` and `global.sniffing_timeout`, propagate it. Accepted forms are unchanged: bare milliseconds, `ms`, and `s`. The new function mirrors its neighbour's shape — a chain of `strip_suffix` early returns — so the two read the same way. Both language references state that anything else is rejected. A deployed configuration carrying one of these values stops loading instead of running with zero, which is the same trade `check_interval` took in #134.

## Verified

The new test is red on the old parser — `check_tolerance: 1m` loads with `check_tolerance_ms: 0` — and green after. `cargo test -p honk-config` passes 173 tests, `cargo fmt --all -- --check` and `cargo clippy --all --all-targets -- -D warnings` are clean, and the workspace gate passes apart from `routing::tests::test_geosite_cn_route`, which fails the same way on an unmodified `45f1a8a` for an unrelated reason.

One limit worth naming: through `Config::from_file` this error is replaced by the structured-format fallback, so a `.dae` file without an `include` reports `Parse error: expected value at line 1 column 1` rather than the setting name. That is pre-existing — `parse_checked_bool` behaves identically today — and I left it alone rather than widen this change.

### DIFF
--- a/crates/honk-config/src/parser/mod.rs
+++ b/crates/honk-config/src/parser/mod.rs
@@ -803,7 +803,7 @@ fn parse_global_section(section: &Section) -> Result<GlobalConfig, crate::Config
         cfg.check_interval_secs = parse_duration_secs(v);
     }
     if let Some(v) = kv.get("check_tolerance") {
-        cfg.check_tolerance_ms = parse_duration_ms(v);
+        cfg.check_tolerance_ms = parse_checked_duration_ms(v, "global.check_tolerance")?;
     }
     if let Some(v) = kv.get("dial_mode") {
         cfg.dial_mode = v.clone();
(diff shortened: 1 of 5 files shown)


Glassyiris:
我们可以直接指定单位为ms 例如 50 就是ms， 在这种场景下很难有其他值类型

Glassyiris:
先合并，后续补默认值

PR #166, opened 2026-09-06
fix(group): keep the URLTest group in a retry plan selection chain

Zakkaus:
## Problem

`urltest_retry_plan_for_target` says it returns entries that "keep the same Honk attribution and selection chain as an ordinary target-aware plan" (`group/score/selection.rs:138-140`). It does not. Direct candidates start with `selection_chain: vec![node.name]` (`:516`), and `selection_plan_for_target` prepends the group at `:355`; the retry builder passes its candidates straight to `score_selection_plan` without that step. A URLTest group `auto` whose first dial fails and whose retry succeeds on leaf `b` therefore reports the chain as `["b"]` where the ordinary plan reports `["b", "auto"]`, which is what Clash `/connections` shows.

## How I fixed it

The retry builder prepends the group name to each candidate's chain, the same insert the ordinary plan already does.

## Verified

The new test builds one URLTest group and compares the two plans for the same target; it fails without the change with `left: "a", right: "auto"` and passes with it. `cargo test -p honk-outbound` passes 653 tests, and `cargo fmt --all -- --check` and `cargo clippy -p honk-outbound --all-targets -- -D warnings` are clean. Attribution is left alone: the ordinary plan prepends the group there only for a Score group, and this builder returns early for any policy but URLTest.

### DIFF
--- a/crates/honk-outbound/src/group/score/selection.rs
+++ b/crates/honk-outbound/src/group/score/selection.rs
@@ -182,6 +182,10 @@ impl super::GroupManager {
             .into_iter()
             .filter(|candidate| seen.insert(candidate.node.id))
             .take(3)
+            .map(|mut candidate| {
+                candidate.selection_chain.insert(0, group.name.as_str());
+                candidate
+            })
             .collect();
         self.score_selection_plan(candidates, super::SelectionPlanMode::Authoritative, context)
     }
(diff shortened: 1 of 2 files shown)


Glassyiris:
reviewed, 当前先合并， 后续 urltest / score 的部分重复时间需要重构

PR #168, opened 2026-09-06
fix(geo): warn when an asset carries a category code twice

Zakkaus:
## Problem

Both geo indexes are built with a plain `HashMap::insert` keyed on the category code — `routing/geo.rs:503` for geosite and `:623` for geoip — and nothing in that file detects a repeat. An asset that carries the same code twice, which concatenated community builds do, silently keeps only the last block and discards the earlier one's entries. Which rules that changes depends on the blocks: a positive condition can match less, a negated one can match more, and identical blocks change nothing. The operator has nothing to look at either way. The module already takes the opposite position elsewhere: a code that expands to nothing warns, with a comment beside it saying never to stay silent about that.

## How I fixed it

Each insert now reports a replaced entry, naming the code and how many domains or networks were dropped. The last-block-wins behaviour is unchanged; whether duplicate blocks should merge is a separate question, and merging would change which rules match rather than only what is logged.

## Verified

A new test builds a geosite fixture of two duplicate blocks plus a single-occurrence control, and a geoip fixture of two duplicate blocks carrying different networks, then captures the log through the same tempfile subscriber `geoip_code_expanding_to_nothing_warns_like_geosite` already uses. Four negative controls were run against it: removing either warning fails its own assertion; making the geosite warning fire for every code fails the quiet-code assertion; and keeping the first geoip block instead of the last fails with `left: [1.0.0.0/8], right: [2.0.0.0/8]`. `cargo test -p honk-core --lib` passes 962 of 963 here; the one failure is `routing::tests::test_geosite_cn_route`, which fails the same way on an unmodified `45f1a8a` because it depends on a sibling test's environment override, and #160 is that fix. `cargo fmt --all -- --check` and `cargo clippy -p honk-core --all-targets -- -D warnings` are clean.

### DIFF
--- a/crates/honk-core/src/routing/geo.rs
+++ b/crates/honk-core/src/routing/geo.rs
@@ -500,7 +500,16 @@ fn parse_geosite_index(
                 }
             }
         }
-        index.insert(code, IndexedGeosite { domains, attrs });
+        if let Some(previous) = index.insert(code.clone(), IndexedGeosite { domains, attrs }) {
+            // Same reason the zero-expansion paths warn: the earlier block's
+            // entries are gone, and which rules that changes is not something
+            // the operator can see from the asset.
+            tracing::warn!(
+                code,
+                dropped = previous.domains.len(),
+                "duplicate geosite code in asset; only the last block is used"
+            );
+        }
     }
 
     Ok(index)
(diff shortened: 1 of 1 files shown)


Glassyiris:
reviewed

PR #182, opened 2026-09-09
fix(config): strip unquoted comments from group filter lines

Zakkaus:
## Problem

`parse_group_section` reads filter lines with `split_once(':')` at `crates/honk-config/src/parser/mod.rs:1005` and never strips a trailing comment, unlike `parse_kv_pair` at `:701`. So `filter: name('edge') # comment` fails `strip_suffix(')')` at `:528`, is dropped by the `filter_map` at `:465`, and — because the emptiness guard at `:452` tests the pre-parse vector — the group silently ends up with no members. `doc/en/configuration.md:26` already says `#` starts an unquoted trailing comment.

## How I fixed it

The filter value now goes through the existing `strip_unquoted_comment` (`:716`). That truncation opens a second hole, so the same commit closes it: `filter: group(hk#suffix)` becomes `group(`, `extract_fn_args` returns `None` (`:878`), the value lands in `group.filters`, `:449` drops it for starting with `group(`, and `:452-459` then hands the group every node. A `group(`-shaped value whose extraction failed now stays in the vector the guard tests, so the group stays empty instead. Retention is limited to failed extraction — `Group.filters` is public and deserializable (`node.rs:339-361`), and a caller supplying a working `group('hk')` must keep reaching the all-nodes path; a test pins that. Three inputs change behaviour, each following `configuration.md:23-26`: `filter: group('hk'` goes from every node to none; `filter: group(hk#suffix)` no longer names a subgroup and yields no members; an unquoted `filter: name(edge#1)` stops matching a node literally named `edge#1`. A quoted `#` inside an argument stays data.

## Verified

`cargo fmt --all -- --check`, `cargo clippy --all --all-targets -- -D warnings` and `cargo test --workspace --no-fail-fast -- --skip test_routing_with_config_dae` all pass. Each hunk has a test that fails without it: `test_group_filter_trailing_comment` fails without the stripping; `test_group_filter_unterminated_group` fails on `main` (the pre-existing hole) and without the retention; `test_group_filter_unquoted_hash` fails on `main` because the comment survives as the subgroup name `hk#suffix`, and fails with stripping alone because the group then takes every node, so it covers the interaction of the two. I also diffed parsed configs against `origin/main` over the three shipped `.dae` files and eight boundary inputs, normalising ids, timestamps and HashMap order: only the three intended cases differ. I did not run the root-only `just test-netns` gate, which this change does not reach.

### DIFF
--- a/crates/honk-config/src/parser/mod.rs
+++ b/crates/honk-config/src/parser/mod.rs
@@ -447,7 +447,8 @@ pub fn resolve_group_filters(groups: &mut [Group], nodes: &[Node], subscriptions
             .filters
             .iter()
             .map(|filter| filter.trim())
-            .filter(|filter| !filter.starts_with("group("))
+            // Unterminated group filters must not trigger the all-nodes fallback.
+            .filter(|filter| !filter.starts_with("group(") || !filter.contains(')'))
             .collect();
 
         if filters.is_empty() {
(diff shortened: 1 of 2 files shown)


Zakkaus:
Run echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | \
KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"
Error: Process completed with exit code 1.
 沒轍 ci的錯

Glassyiris (commented):
clear

PR #186, opened 2026-09-09
feat(config): report the parser's lenient fallbacks as diagnostics

Zakkaus:
## Problem

#184 gave the parser a diagnostics channel and wired one producer, the timer fallback. Every other lenient site still takes a fallback silently: `tproxy_port` and `pprof_port` (`crates/honk-config/src/parser/mod.rs:772,778`), `so_mark_from_dae` (`:781`), `check_interval` and a subscription's `interval` (`:833,1133`), `ipversion_prefer`, `optimistic_cache_ttl`, `max_cache_size` and a bad `fixed_domain_ttl` entry (`parser/dns.rs:37,43,46,233-236`), the eleven `parse_bool` callers (`mod.rs:1250`, anything outside `true/yes/1/on` is `false`), an unknown group policy (`:1091`, documented as silently `selector`), and a `filter:` line that honk cannot parse (`:490-493`, dropped, which can leave a group with no nodes). A user who mistypes one of these gets the fallback and no trace of it.

## How I fixed it

Three commits. The first two change no `Config` for any input; the third is the one behaviour change Glassyiris asked for while reviewing: `ipversion_prefer: 0` and a value honk cannot parse now mean no preference (`Both`, the omission default), not prefer-IPv4.

`27248fe` adds one helper, `lenient(parsed, fallback, diagnostics, || diagnostic)`, and moves every scalar site onto it; `lenient_duration_ms` becomes a call to it with its message unchanged. Booleans go through `lenient_bool`, which recognises dae's twelve spellings (`t`/`f`/`y`/`n` included) without a diagnostic and reports anything else; `t` and `y` still yield `false`, as before. `parse_bool` and the local `parse_duration_secs` are deleted; `parse_hex_or_dec` and `parse_ip_prefer` return `Option` so their callers can report, with the same parsing (`0`, `4`, `6` accepted, hex-first mark unchanged). The unknown-policy arm reports the group by name with an empty value, because policy text is free-form. Messages say what honk could not parse, never that the value is invalid: `0x3039`, `1.5s` and `0x6` are valid dae that honk does not accept, and the diagnostic now says so. Settings are qualified like the existing ones (`global.tproxy_port`, `dns.optimistic_cache`, `subscription.<name>.interval`, ...). `diagnostics` is threaded into the dns, group, subscription and experimental section parsers; the node parser and every public signature are untouched. `groups.md` in both languages no longer says "silently".

`d6fb786` reports a dropped group filter by its ordinal among the group's node filters (`group(...)` declarations are not counted, they never reach `group.filters`), with a fixed message and no filter text, since a filter can carry a share link. `resolve_group_filters` keeps its signature and stays silent: honk-core re-resolves at startup (`honk-core/src/lib.rs:679`) after the parse-time diagnostics were already reported (`:618`), and again on SIGHUP and subscription rebuilds, so a logging wrapper would repeat every warning. The parser calls a private `resolve_group_filters_inner` with a `Some` sink; the runtime path passes `None` and builds nothing.

## Verified

`cargo fmt --all -- --check`, `cargo clippy --all --all-targets -- -D warnings`, `cargo test -p honk-config` (140 + 67) and `cargo test --workspace --no-fail-fast -- --skip test_routing_with_config_dae` (1974 passed, 0 failed) pass. Every production hunk has a test that fails without it: each scalar, policy and boolean site reversed alone fails its parser test on the diagnostic count (0 vs 1; 10 vs 11 for each of the eleven boolean sites); the filter push reversed fails five parser tests and the extended daemon lifecycle test, which runs the real binary with `--mock-ebpf`, counts exactly one `group.<name>.filter` warning by `Router ready` and exactly two after `SIGHUP reload request 1 applied`, so the runtime paths do not duplicate it. A differential run over 88 dae inputs (the repository samples, every raw-string literal in `parser/tests.rs`, and one file per site including `0x3039`, `1.5s`, `+6`, `06`, `0x6`, `4294967296`, `t`/`f`/`y`/`n`, `name('a&&b')`, an ignored nested `cache` block, and filters and policies carrying a fake credential) shows byte-identical `Config` output between `1dfa985` and this branch, exactly 37 added diagnostics matching a manifest written before the branch existed, and no credential in any diagnostic. Not run: the eBPF job.

Recorded for a separate change, not touched here: `so_mark_from_dae: 10` reads as 16 (dae reads 10); `allow_insecure: t` is true in dae and false here; `check_interval: 1.5s`, `tproxy_port: 0x3039` and `name('a&&b')` are valid dae that honk cannot parse; and a subscription line without a tag (dae's own `example.dae:166`) is split at the wrong colon.

`3818c3e` (added after review): `parse_ip_prefer` maps `0` to `DnsStrategy::Both` and the lenient fallback is `Both` with the message "using fallback: no preference"; three test expectations and the `ipversion_prefer` rows of both references change with it. In the differential corpus exactly `ipversion_prefer: 0` and `ipversion_prefer: ipv4` change, to `Both`. Gates rerun on the branch head: fmt, clippy, honk-config 207, honk-core `dns::` 350.


Glassyiris:
clear

PR #188, opened 2026-09-09
fix(config): treat ipversion_prefer 0 and unparseable values as no preference

Zakkaus:
Stacked on #186 (its two commits appear here until it merges).

## Problem

Glassyiris pointed out on #186 that an omitted `ipversion_prefer` means `both` (`crates/honk-config/src/dns.rs:760`), while the parser mapped `0` and any value it could not parse to `PreferIpv4` (`parser/mod.rs:1403-1407` on the #186 head). dae defines `0` as "no preference", which is the same thing as omitting the setting. #186 deliberately kept that mapping and only reported it; this is the behaviour fix.

## How I fixed it

`parse_ip_prefer` maps `0` to `DnsStrategy::Both`; the lenient fallback for an unparseable value is `Both` and the diagnostic says "using fallback: no preference". Three test expectations and the `ipversion_prefer` rows of the English and Chinese references change with it. Nothing else: `4` and `6` keep their prefer variants, and every honk-core consumer already treats `Both` as the default strategy (request filtering, preference suppression, singleflight metadata, policy identity all handle it).

## Verified

`cargo fmt --all -- --check`, `cargo clippy --all --all-targets -- -D warnings`, `cargo test -p honk-config` (207), `cargo test -p honk-core --lib dns::` (350) pass. Behaviour change for exactly two inputs in the #186 differential corpus: `ipversion_prefer: 0` and `ipversion_prefer: ipv4` now yield `Both` instead of `PreferIpv4`; the diagnostics are unchanged apart from the wording.


Zakkaus:
Folded into #186 at the maintainer's request.

PR #189, opened 2026-09-09
refactor(config): scan blocks once and split expressions outside quotes

Zakkaus:
## Problem

`honk-config` reads a dae file with eleven independent line scanners, each with its own idea of where a brace, a quote or a comment ends. Braces inside quoted values, one-line blocks, a statement on the same line as a nested `}`, and a `,`, `)`, `&&` or `->` inside a quoted matcher argument are all split in the wrong place.

## How I fixed it

One block scanner (`parser/structure.rs`) now reads the brace structure once: whole-line `#` comments, quoted spans protecting braces, an unmatched `}` reported as a diagnostic and an unclosed block as an error naming its opening line. Section readers consume the block tree through a projection that keeps today's per-section behaviour for unknown nested blocks, so the differential for every existing input is unchanged apart from the classes listed in the docs. The named expression sites (`extract_fn_args`, `strip_tag_arg`, the routing and DNS `->` and `&&` splits, the DNS comment rule, the `group(...)` guard) share two quote-aware helpers instead of `find`/`split`. Three commits: scanner, cutover, quote-aware sites. dae's `[add_latency: …]` filter annotation is deliberately not read: honk has no such feature, so the filter stays on the existing "could not parse this filter; ignored" diagnostic instead of being silently accepted with a different meaning.

## Verified

`cargo fmt --check`, `cargo clippy --all --all-targets -- -D warnings`, `cargo test -p honk-config` and the workspace gate pass on every commit. A frozen corpus of 211 configurations (repository samples, every test literal, the documentation examples, the dae upstream example, an include graph and the boundary fixtures) was dumped in four modes at each commit and compared byte for byte against expectations written before implementation: 0 of 844 differ at each of the three commits. The statement sequence produced by the old scanners and the new projection is identical for 182 of 211 files; the 29 others are include probes and the structural classes the change is for. Reversing the production hunks of each commit fails exactly the expected cells and nothing else; the sixteen quote-aware sites were reversed one at a time and each breaks its own witness. Not run: `just test-routing` and `just test-netns`.

---


Glassyiris:
clear， tested on vyos

PR #190, opened 2026-09-09
fix(config): read subscription entries with dae's tag rule

Zakkaus:
## Problem

The `subscription` reader splits every line at its first `:`, so dae's own `example.dae` lines `'https://example.com/no_tag_link'` and `'file://…' # comment` become name `'https` with url `//example.com/no_tag_link'`, and `Config::validate` then rejects the whole file. Trailing comments on node and subscription lines are kept as data, and a User-Agent suffix followed by a comment (`sub: 'http://sub'(honk/1.0 like) #xxxx`) is not recognised at all.

## How I fixed it

One helper decides the tag for node and subscription lines the way dae does: the text before the first `:` is the tag unless that colon starts `://`, quoted tags and links are found with the escape-aware quote scan the block scanner already uses, and each reader keeps its own normalisation, so node output is unchanged except for tags containing an escaped quote. A tagless subscription takes its URL host as its name (`example.com`); a line with neither a tag nor `://` is ignored as before. A `#` after a space or tab ends a node or subscription line; a glued `#` stays part of the link. Inside a quoted subscription value a `#` right after the closing quote or after the `)` that closes the `(UA)` suffix is also a comment, so the maintainer's `sub: 'http://sub'(honk/1.0 like) #xxxx` parses with the UA. Three commits: the shared tag split, tagless entries, comments.

## Verified

`cargo fmt --check`, `cargo clippy --all --all-targets -- -D warnings`, `cargo test -p honk-config` and the workspace gate pass on every commit. The 211-file parser corpus plus 19 new node and subscription fixtures were dumped in four modes at each commit and compared byte for byte against expectations written before implementation: 0 of 920 cells differ at each commit, and reversing each commit's production change fails exactly the cells that commit is meant to change. New tests assert exact names, urls, user agents and node endpoints for every changed and every pinned shape, including validation of an empty host name and `subtag` membership for a name collision. Not run: `just test-routing`, `just test-netns`.

---


Glassyiris:
clear

PR #192, opened 2026-09-10
perf(routing): reuse repeated fact lookups lazily

Glassyiris:
## Problem

Repeated destination IP, source IP, and MAC conditions look up the same immutable generation facts again even when they only need different bitmap bits. Eliminating that work must preserve ordered first-match routing and the complete decision.

## How I fixed it

Cache each fact category lazily only when it appears at least twice, retaining both hits and NULL results in invocation-local stack slots. Separate full-width readiness slots preserve Linux 6.12 verifier precision; domain entry lookup, condition negation, ABI, plan identity, and publication ordering stay unchanged. Add semantic/root-slot regression coverage and update the bilingual routing design docs.

## Verified

Passed workspace tests (1,988), fmt, workspace and eBPF-specific clippy, test-routing/netns, and real x86/ARM kernel checks including Linux 6.12. Five paired LAN/WAN L2/L3 TCP/UDP IPv4/IPv6 lab runs per dataset covered direct/proxy/block/DNS and new/cached flows, with both hosts restored. [Pinned review, commands, and raw evidence](https://github.com/Glassyiris/honk/blob/8fa2bfe35a185c6b6c8a1f361b5aaa2dda499b1d/bench/results/routing-fact-cse-2026-09-10/result.json) separate routing-component gains from mixed end-to-end latency and increased worst-case verifier/JIT cost; lab binaries were manual musl builds, not release artifacts.

### DIFF
--- a/crates/honk-core/src/control/routing_matcher/codegen.rs
+++ b/crates/honk-core/src/control/routing_matcher/codegen.rs
@@ -48,6 +48,50 @@ const MUST: i16 = std::mem::offset_of!(RoutingDecision, must) as i16;
 const DOMAIN_FINAL: i16 = std::mem::offset_of!(RoutingDecision, domain_final) as i16;
 const RULE_ID: i16 = std::mem::offset_of!(RoutingDecision, rule_id) as i16;
 
+#[derive(Clone, Copy)]
+enum FactKind {
+    Destination,
+    Source,
+    Mac,
+}
+
+impl FactKind {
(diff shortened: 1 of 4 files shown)


Glassyiris (commented):
clear, need test

reviewed by  @Glassyiris 

design/code: @Glassyiris 
test/ci/bench: GPT-6-ASTRA Xhigh

PR #195, opened 2026-09-10
docs(config): describe the honk dialect of dae configuration

Zakkaus:
## Problem

honk reads dae's syntax as a dialect, and the maintainers have decided not to align the differences with dae. Nothing documents them: a reader moving a file between dae and honk finds out by loading it, and several divergences change what the engine enforces (a glued `#` cut from a scalar, `dport (443)` dropping its condition, a DNS action lowercased, a subscription URL carrying its comment).

## How I fixed it

A new reference page, `doc/en/reference/dialect.md` with a `doc/zh/` mirror, lists every known difference in seven tables (lexical rules, scalars and lists, node and subscription entries, routing, DNS, groups, sections and includes). The dae column states only what `dae_config.g4` says; the honk column is what the parser does on `66fc946`, produced by loading each example through `parse_dae_config`, so any row can be re-checked by loading the same text. The configuration guide's opening sentence links to it in both languages. No code changes.

## Verified

Every honk-column example was dumped through the parser on `487e942` and `66fc946` and the text written from those dumps; the page went through five review rounds against the grammar and the parser source, which removed guessed dae semantics, split lenient from strict booleans, and added rows for one-line blocks, matcher spacing, upstream comments, non-ID block names and string-parsed includes. Chinese wording passed the project's lint. Not run: anything else; documentation only.

---


Glassyiris (commented):
clear

Issue #194, opened 2026-09-10
ConnectionPool: a checkout racing purge_node takes the purged stream and double-subtracts total_entries


Zakkaus:
`ConnectionPool::acquire_entry` (`crates/honk-core/src/pool.rs:259-282` on `487e942`) first clones the entry's `Arc` out of the map, then locks the vector:

```rust
let arc = Arc::clone(&*self.entries.get(addr)?);
let mut list = arc.lock();
```

`purge_node` (`:469-482`) removes map entries with `retain`, counting each vector's length under a brief lock, then subtracts the sum from `total_entries` once.

Interleaving, on two workers:

1. checkout clones the `Arc` for `ready|<addr>|<target>` and blocks on `arc.lock()` while `retain` holds it;
2. `retain` reads `len() == 1`, drops the map entry, and `purge_node` subtracts 1;
3. checkout acquires the lock on the vector it still holds, finds the live stream, `swap_remove`s it, and subtracts 1 again (`:281`).

Result: the purged stream — the whole point of `purge_node` is that a dead node's stream must never be handed out — is handed out, and `total_entries` goes from 0 to `u64::MAX`, after which `reserve_total` refuses every deposit until the process restarts.

The only production caller of `purge_node` is the health-death callback (`control/bootstrap.rs:314-332`), so the window opens exactly when a node is declared dead while a flow is checking out one of its streams. `purge_bare` (`:450-462`) has the same shape for bare entries.

A fix is either to make removal drain the vector under its lock and let the checkout observe an empty vector, with the count adjusted once per stream actually removed, or to take the entry out of the map inside the same critical section the checkout uses. Found while planning #145.

Zakkaus:
Fixed by #196 (first commit): all pool state now lives under one lock, so a checkout, a deposit, a purge and the janitor cannot interleave on a key, and the total is exact by construction. The four interleavings are pinned by pause-hook tests that fail on the old structure.

Zakkaus:
Fixed on `main` by #196: `ConnectionPool` now keeps its entries, total, ready-target counts, warm claims and hotness in one `Mutex<PoolState>`, so a checkout, a deposit, a purge and the janitor cannot interleave on a key and the total is exact by construction. The four interleavings from this report are pinned by pause-hook tests that fail against the old structure.

Issue #145, opened 2026-09-06
pooled ready streams cross credential and retirement boundaries


Zakkaus:
Two independent mechanisms let a pooled, already-authenticated stream cross the identity boundary it was created under.

### 1. The ready key does not include credentials

`crates/honk-core/src/pool.rs:225`, on `becdbd3`:

```rust
/// Pool key for a Ready entry. The completed handshake bound the
/// stream to (proxy node, target), so the key contains both; with
/// domain routing the CONNECT request carries the domain, making
/// `domain:port` — not the resolved IP — the destination identity.
pub(crate) fn ready_key(
    node_addr: &str,
    target: SocketAddr,
    target_domain: Option<&str>,
) -> String {
```

The comment says the key contains the proxy node. The parameter is `node_addr`, the proxy's `host:port`. Two nodes that share an address and differ only in credentials produce the same key, and the second one takes the first one's completed handshake.

Reproduced earlier against a local SOCKS5 origin: with node A and node B on one address and different username/password, switching the selector from A to B moved `readyHits` from 0 to 1 and B's request travelled on A's authenticated connection. The peer log ordering confirms it — B's payload appears before any auth for B, because there was none. Changing the destination port instead grew `readyMisses` normally, so destination isolation works and credential isolation is the missing dimension.

The namespacing test at `pool.rs:784` varies only the address (`"proxy.example:1080"` vs `"other.example:1080"`), which is why this has not shown up: the credential axis is not covered rather than deliberately accepted.

### 2. Retirement does not evict ready streams

`crates/honk-outbound/src/runtime.rs:1002`:

```rust
pub fn begin_retirement(&self) {
    self.terminal.store(true, Ordering::Release);
}
```

That marks the generation terminal. `retire_reusable_state()` retires generation-owned AnyTLS/VLESS/QUIC resources, not core's `ConnectionPool`, and selector reconciliation calls `purge_bare`, which `pool.rs:448` explicitly preserves ready streams through. The acquisition path at `crates/honk-core/src/control/connection/tcp.rs:1109` checks entry kind, expiry and socket liveness — not retirement.

Full eviction already exists at `pool.rs:469` (`purge_node`); its only production caller is the health-death callback in `control/bootstrap.rs:320`, not node retirement.

[...]

PR #196, opened 2026-09-10
fix(pool): one lock for the connection pool, and ready streams that stay with their generation

Zakkaus:
## Problem

`ConnectionPool` keeps its streams, its total, the per-node ready-target counts, the warm-dial claims and the hotness map in five separately synchronised structures, so the invariants the code assumes do not hold under concurrency: a checkout racing a health purge can take the purged stream and subtract twice, a deposit can land in a vector the checkout has just unlinked and vanish with its reservation, the janitor's snapshot store loses a concurrent deposit, and once the total is wrong the pool refuses every deposit until restart (#194). Separately, ready streams are keyed by the proxy's `host:port`, so two nodes on one address with different credentials share each other's authenticated streams, and a reload leaves the old node's ready streams reachable until they expire (#145).

## How I fixed it

The first commit puts all pool state under one `parking_lot::Mutex<PoolState>`: every operation takes it once, never across an `.await`, never touches a second key inside it, and evicted streams are destroyed after it is released. The total is then exact by construction, a `check_invariants` helper states the invariants and the tests call it after every step, and the four races above are reproduced by pause-hook tests that fail on the old structure. The second commit keys ready streams by runtime generation and node identity (`ready|<generation>|<id>|<target>`; `OutboundRuntimeRegistry` gets a generation number) so a re-credentialed or replaced node can never find its predecessor's streams, and retires a generation as one pool operation right after the reload's `begin_retirement`: its ready streams, target budget, warm claims and hotness are removed and late deposits, hotness updates and warm claims for it are refused. Health purges narrow from the whole address to the identity that died. Limits, bare-TCP keys, selector warm ownership and `derive_id` are unchanged.

## Verified

`cargo fmt --check`, `cargo clippy --all --all-targets -- -D warnings`, `cargo test -p honk-core`, the workspace gate and `just outbound-ci` pass on both commits. Each failing-first test was run against the previous structure and failed by assertion: the four race tests, the credential and dial-shape isolation tests against a loopback RFC 1929 SOCKS5 server, the two-generation and per-identity budget tests, the retirement and late-write tests, the 70-generation hotness test, and the three reload cases in `reload_tests` (re-credentialed, replaced, unchanged-plus-routing-change; a rejected reload keeps its stream). Eleven single-hunk ablations each make their test fail. A non-gating loopback measurement with 2,048 pooled streams put a janitor pass at about 250 µs and a full generation retirement at about 2 ms under the lock. Not run: `just test-netns`, the eBPF job.

---

### DIFF
--- a/crates/honk-core/src/control/bootstrap.rs
+++ b/crates/honk-core/src/control/bootstrap.rs
@@ -315,17 +315,20 @@ impl ControlPlane {
         let pool = self.connection_pool.clone();
         let udp_pool = self.udp_pool.clone();
         let config_for_purge = self.config.clone();
+        let runtime_registry = self.runtime_registry.clone();
         self.alive_set.set_death_callback(Some(Box::new(
             move |node_id: uuid::Uuid, _name: &str| {
                 udp_pool.remove_by_node(node_id);
                 let node_addr = config_for_purge.try_read().ok().and_then(|c| {
-                    c.nodes
-                        .iter()
-                        .find(|n| n.id == node_id)
-                        .map(|n| format!("{}:{}", n.host(), n.port))
+                    c.nodes.iter().find(|n| n.id == node_id).map(|n| {
+                        (
+                            format!("{}:{}", n.host(), n.port),
+                            runtime_registry.read().generation(),
+                        )
+                    })
                 });
-                if let Some(addr) = node_addr {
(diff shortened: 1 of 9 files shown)


Glassyiris (commented):
reviewed.

PR #199, opened 2026-09-10
fix(dns): negative answers take their lifetime from the answer, not a default

Zakkaus:
## Problem

DNS lifetime handling mixed cache admission, upstream response semantics, and routing projection: negative or zero-TTL answers could be reused incorrectly, response rejection could select the wrong TTL policy, and uncacheable positive answers could lose their domain-routing facts.

## How I fixed it

Honor fixed-domain TTL zero for every response code, use usable SOA lifetimes for NXDOMAIN/NODATA, and supersede unusable answers under the existing cache-enable, epoch, and refresh-revision guards. Select TTL policy from the original response even when rejection generates synthetic NOERROR. Keep routing lifetime independent of cache admission, handle valid 11-byte resource records, consolidate cache publication, and share projection-test fixtures. Preserve SERVFAIL behavior, configured overrides, and existing persistence limits; update both documentation languages.

## Verified

The code committed as d71bb2f passed `ci/dns-ci.sh` (including fmt/clippy) and `cargo test --workspace --no-fail-fast -- --skip test_routing_with_config_dae`: 2077 passed, 4 ignored, 1 filtered. Focused suites passed 88 forwarder and 5 controller UDP tests; failing-first regressions cover rejected REFUSED, minimum-size records, and projection lifetime. Tests used `CARGO_PROFILE_TEST_DEBUG=0` and one build job after compiler SIGKILLs. Real eBPF/netns tests and benchmarks were not rerun; these local results are not a claim that new-head GitHub CI has passed.


Zakkaus:
One interaction to flag on my own PR, found while scanning the DNS controller after this was opened.

`DnsController::submit_projection` (`control/dns_control.rs:317`) uses the cache expiry as the routing lifetime: `advertised_ttl: outcome.expiry().ttl()`. An accepted positive answer with `EffectiveExpiry::do_not_cache()` therefore installs a domain owner that is already expired, and the projection worker removes it before it reaches the domain map. That is pre-existing: `use_host` answers and `fixed_domain_ttl: 0` names already behave this way on `main` (reproduced with tests against `MockEbpfBackend`; I will file it separately).

Commit 3 of this PR widens it by one case. Before, a positive rcode 0 answer whose records all carry TTL 0 was cached for 1 s under `optimistic_cache_ttl: 0` (`answer_ttl.max(1)`), so domain routing learned its addresses for that second. After commit 3 it is not cached, its projection lifetime is zero, and `domain(...)` rules no longer see it at all. The cache decision in commit 3 is right; the routing lifetime should not be borrowed from it. The fix belongs in the controller (derive the projection lifetime from the answer the client actually receives, with whatever floor domain routing needs), not in the cache admission.

I can either add that controller change here or leave commit 3 out of this PR until it lands separately; your call.

Glassyiris (approved):

PR #216, opened 2026-09-10
fix(udp): preserve original reply tuples for domain targets

Glassyiris:
## Problem

A proxy can resolve a UDP domain target to a different address than the client's original destination. The endpoint's [reply-source socket selection](https://github.com/daeuniverse/honk/blob/7812000d8ea68f2aeb81be8f2f5a31503d0aa5a6/crates/honk-core/src/control/udp_endpoint/driver.rs#L664-L712) exposed that address from VLESS XUDP/Mux.Cool metadata to the client, breaking the original reply tuple and causing HTTP/3 timeouts.

## How I fixed it

Carry the canonical UDP dial target's domain flag into the endpoint. After the existing transport-peer validation, map accepted domain-target replies back to the original destination IP, port, and address family; literal-IP full-cone replies retain their existing source handling. Extend the existing reply-source regression with explicit literal/domain cases sharing one real-driver lifecycle, preserving connected-client, B/A/B source, cross-family and cache-reuse checks, and update the English and Chinese control-plane documentation. The concurrent CI update retains real-kernel testing through software emulation when KVM is unavailable.

## Verified

The consolidated regression passes normally and rejects both missing and unconditional reply remapping in isolated negative controls. All 51 UDP endpoint tests and the workspace gate passed (2,057 passed, four ignored, one known legacy routing test skipped), as did workspace Clippy with `-D warnings`; local tests used one build job and `CARGO_PROFILE_TEST_DEBUG=0`. Five additional isolated driver scenarios passed: changed source port, IPv4 replies to an IPv6 client, empty datagrams, unexpected fixed-peer rejection, and transport-error non-delivery. Built the eBPF-enabled x86_64 musl binary and manually deployed these production sources via scp/sudo to a VyOS 6.18.33 gateway; Google IPv4 and YouTube IPv4/IPv6 returned HTTP/3 200, with TCP/DNS smoke checks also completed. That prior deployment used the unchanged production implementation, not a release artifact (binary SHA-256: `fad70832cd974e184206558fd147bb9235de38f5dea8423796e31a8bcfcd2e9a`). `just test-routing`, `just test-netns`, and a full live protocol/kernel matrix were not run.


Glassyiris (commented):
clear

PR #227, opened 2026-09-10
fix(config): ship a loopback controller, and reject dns upstreams no query can resolve

### LINKED ISSUE #221: shipped config.dae binds the clash api to 0.0.0.0 with an empty secret
`config.dae:226-232` ships:

```dae
experimental {
    clash_api {
        external_controller: '0.0.0.0:9090'
        external_ui: 'yacd'
        secret: ''
```

`doc/en/reference/api.md:17` says the opposite: "The API has no TLS. Bind it to localhost or place a TLS reverse proxy in front of it, and set a strong `secret` when untrusted clients can reach the listener." `:15` adds that an empty `secret` disables authentication.

An operator who starts from the shipped file therefore exposes an unauthenticated control API on every interface, including the LAN the gateway serves. It can switch a selector (`PUT /proxies/{name}`), close connections (`DELETE /connections/{id}`), flush the DNS and FakeIP caches, and read every connection's destination. Nothing warns at startup about the non-loopback plus empty-secret pair.

Fix: ship `127.0.0.1:9090` in `config.dae`, and warn at startup when `external_controller` is not a loopback address and `secret` is empty.

Verified by reading the shipped file and the reference page on `7812000`; the endpoint list is from `crates/honk-core/src/clash_api.rs`.

### LINKED ISSUE #223: dns accepts upstream names that every query then fails to resolve
A named DNS upstream without a request fallback, a misspelled upstream action, or a case mismatch passes validation but fails during resolution.

Sites on `7812000`: `crates/honk-core/src/dns/routing.rs:520-536`, `crates/honk-core/src/dns/planner.rs:170-176`, `crates/honk-core/src/dns/upstream_pool/query.rs:333-336`; validation: `crates/honk-config/src/config.rs:774-794`.

Trigger: configure `dns { upstream { alidns: 'udp://223.5.5.5:53' } }` without request routing. The implicit fallback remains `default`, which the explicit upstream block removed. Queries fail with `unknown upstream: default`; explicitly setting `fallback: alidsn`, or declaring `AliDNS` and using that action, instead fails on `alidsn` or lowercased `alidns`.

Fix: validate actions and the effective fallback against declared upstreams, seed the planner from those declarations, and apply one consistent name-matching rule.

Verified: an executed parser/validator/resolver probe accepted all three configurations and returned the corresponding `unknown upstream` errors; no live gateway was exercised.

Zakkaus:
## Problem

Two operator-facing defects from the whole-project scan, both reachable from a file honk ships or accepts.

`config.dae` bound the Clash control API to `0.0.0.0:9090` with an empty `secret`, while `doc/en/reference/api.md` tells the operator to keep it on loopback and set a secret because the API has no TLS, and `doc/en/configuration.md` repeats that. The API is not read-only: it switches selectors, closes connections, flushes caches and lists every destination. Nothing warned when a non-loopback controller had no secret (#221).

A DNS configuration could name an upstream that no query can resolve and still load: declaring an explicit `upstream` block removes the implicit `default`, so a configuration with no fallback fails every query with `unknown upstream: default`; a typo in an action name, and a response rule whose re-query names an undeclared upstream, do the same (#223).

## How I fixed it

Two commits.

1. `fix(config): ship a loopback controller and warn when an exposed one has no secret`. `config.dae` ships `127.0.0.1:9090`, matching `example.dae` and both configuration pages. `api_exposed_without_secret` decides whether a bound address plus secret is an exposed unauthenticated listener, and the startup path warns once, naming the address, when it is; the server still starts, because an operator who wants that must be able to have it. `crates/honk-config/tests/example_configs.rs` now asserts that the shipped controller parses to a loopback address instead of pinning a literal string. Both reference pages gain the sentence.
2. `fix(config): reject dns upstream names that no query can resolve`. Validation walks the effective request and response routing and rejects any action, and any effective fallback, that names an undeclared upstream, listing the declared ones. Legacy structured configurations are reported at the path the operator wrote (`dns.routing.rules[i].upstream`, `dns.routing.fallback`), not at the synthesised effective path; an omitted or empty legacy fallback says what is actually wrong rather than naming a phantom upstream. `doc/*/reference/dns.md` explains that an explicit block replaces the implicit `default`.

## Verified

- Per commit: `cargo fmt --all -- --check`, `cargo clippy --all --all-targets -- -D warnings`, `cargo test --workspace --no-fail-fast -- --skip test_routing_with_config_dae`: exit 0 (2135 and 2150 tests passed, 0 failed).
- Failing-first for every behaviour: the warning matrix, the captured warning itself, the shipped-bind assertion, and one test per rejected DNS shape.
- Reviewer ablation, one site at a time with the tests kept: inverting the loopback test, inverting the empty-secret test, and suppressing the warning each fail the warning tests; reverting the shipped bind fails `test_config_dae_parses`; disabling the fallback check fails two validation tests. All by assertion, no compile errors.
- Differential: the frozen parser corpus (230 files x 4 modes = 920 cells) is byte-identical to `origin/main`; `config.dae`, `example.dae` and every `.dae` under `doc/` still load.
- A cold reader who had not seen the plans checked which configurations that load today now fail: four shapes refuse, and all four already failed at query time with `unknown upstream`. Its findings about the error text are folded into commit 2.
- `chinese_lint.py` clean on the changed Chinese pages.
- Not run: `just test-routing`, `just test-netns`, the eBPF job.

Closes #221. Closes #223.

---


Glassyiris:
## Deployment verification

AI-assisted deployment performed at the maintainer's request. Deployed merged `main` commit `d8c282cb5f11f9d58966f0bc2aab80134075f220` to the authorized x86_64 VyOS gateway (Linux `6.18.33-vyos`) using the existing `dae.service` and unchanged production configuration. This was a manual source build, not a release tarball: static musl, default mimalloc/rprx/Clash API plus real eBPF; version `v0.0.1.beta.76-7-gd8c282c`, SHA-256 `b41dee2731f1d01fe6cbccb978d455a163926d4c19afd2892fdf5826e8701c2e`.

The main workspace, lint, and real-kernel CI checks passed. The exact shipping binary passed the local DNS process smoke. Production configuration passed the current parser/validator before downtime; the previous binary/configuration and quiesced SQLite database were backed up, and an independent rollback timer protected the cutover. Service interruption was approximately 7.8 seconds. Configuration hashes and the `UDP_DECISION_SEQUENCE` pin inode were preserved; no map-cleaning procedure was used.

Post-deployment: 21 samples over five continuous minutes showed a stable PID, zero automatic restarts, RSS 77.0–91.8 MiB, and zero NFQUEUE verdict/token/kernel-drop errors. Kernel/namespace/map diagnostics passed, direct and proxy UDP decisions were observed, LAN UDP/TCP DNS and IPv4/IPv6 HTTPS succeeded, and unchanged SIGHUP applied without restarting the process. All 34 real proxy leaves returned successful proxied URL-test results across initial/retry probes: four German nodes failed the initial 2.5-second-deadline batch and passed subsequent checks with a 10-second deadline (about 1.6 seconds total each). SQLite quick_check returned `ok`; no service/kernel ERROR, panic, fatal, or OOM entries matched the deployment window.

The rollback timer was disarmed only after verification; rollback binaries and detailed evidence remain on the gateway. Retained operational warnings were an empty `geoip:telegram` expansion, the legacy `.sub` storage location, and an intermittent REALITY handshake EOF followed by successful endpoint checks. Configuration was not changed to suppress them. The compiler also reported two nonexistent BoringSSL library search directories, but linking completed and ELF inspection confirmed no dynamic section/interpreter. No exhaustive protocol interoperability or long-duration load claim is made.

PR #233, opened 2026-09-12
config: rebuild the dae parser on a lexer and a dispenser

Zakkaus:
## Problem

`honk-config` split delimiter handling between `parser::structure` and section readers, giving lexical decisions and diagnostic attribution multiple owners. Active configuration diagnostics also need provider-aware ownership across full reloads, SIGHUP and subscription refreshes.

## How I fixed it

Unified lexical ownership in the source-backed cursor and span readers, preserving compact-block, quoted/bare-value, include-fragment and ignored-line boundaries, expression continuation and diagnostic provenance. Active `/configs` diagnostics use provider-aware publication with duplicate rejection and the existing configuration-before-diagnostics lock order. Updated regression coverage, bilingual references, the module map and reload benchmark. LLM-generated/edited areas include parser/diagnostic code, tests, benchmark/CI, docs and this text; tools were Oh My Pi coding, reviewer and scout agents. Reserved reload work was explicitly authorized under CONTRIBUTING §11.

## Verified

Final source at `e0ef2a9`: workspace **2381 passed, 4 ignored**, with `--skip test_routing_with_config_dae`; honk-config **453 passed**; fmt-check, workspace clippy (`-D warnings`) and `sh ci/dns-ci.sh` passed. The latest hostile review found and fixed quoted-entry hash tails importing another node; its node/subscription regression failed before the guard and passed afterward. The repair delta has no remaining scoped findings, not blanket approval of earlier PR changes. Remote documentation commit `e8f5226` is preserved; rebasing changed no tested source/test bytes. Real-kernel/netns and final-diff performance gates were not rerun locally; human maintainer approval remains required.


Glassyiris (approved):

Zakkaus:
Checked `34f0775d` on top of my tip: fmt and clippy clean, workspace gate 2367 passed / 0 failed / 4 ignored, and both frozen corpora (844 + 920 cells, string/parsefile/file/resolve) produce byte-identical Config and diagnostic sections against `fe060415`, so the boundary and ownership changes alter no observed parse output on those inputs.

One doc line was behind the code: the dialect table's tag row still said "the text before the first `:` is the tag", while `split_entry` (`entries.rs:460-474`) only accepts a colon right after the closing quote for a quoted head, as `subscription.md` already states. Pushed `e8f5226c` to bring the en and zh rows in line.

One question, your call: `reload_runtime_config` and `merge_subscription_nodes` now take the diagnostic argument directly and every in-tree caller is migrated. The plan had promised the old signatures for one release behind wrappers; if you want that kept I can add the two wrappers, otherwise I will record the new signatures as the contract.

PR #234, opened 2026-09-12
fix(probes): share HTTP measurement across health checks and URLTest

Glassyiris:
## Problem

Periodic health checks and URLTest maintain separate HTTP exchanges, and URLTest still discards the configured request path ([health implementation](https://github.com/daeuniverse/honk/blob/27426b2cbda63bd02b405326c59478d04109071c/crates/honk-core/src/control/probers.rs#L301-L434), [URLTest request](https://github.com/daeuniverse/honk/blob/27426b2cbda63bd02b405326c59478d04109071c/crates/honk-outbound/src/urltest.rs#L599)). Their response validation, fallback and temporary-runtime cleanup can therefore diverge.

## How I fixed it

Share outbound HTTP execution and explicit generation warm-up, leaving core as an adapter and keeping health/Score/delay accounting separate. Reuse main's canonical check-target decoder and fallible node admission, preserve sanitized request targets, and tighten response bounds, H2 shutdown and fallback behavior. Update bilingual documentation. All changes to probe source, regression tests and documentation/AGENTS were LLM-assisted using OMP task agents; omo-review is advisory, not human approval.

## Verified

Passed outbound-ci, workspace clippy with all targets and warnings denied, and the workspace gate: 2286 passed, 4 ignored, with only the documented test_routing_with_config_dae exclusion. Also passed 30 focused URLTest tests, 4 core probe tests and public-API loopback smoke covering raw dot/query targets, fragments and credential-free native requests; rebase integration received five-lane/hostile review. The explicitly accepted upstream P2 [hyperium/h2#958](https://github.com/hyperium/h2/issues/958) remains deferred without a fork/vendor patch. Real eBPF/netns, musl and live proxy-server matrices were not run; human maintainer approval is still required.


crates/honk-outbound/src/urltest.rs
```diff
@@ -497,160 +526,321 @@ fn https_connector() -> anyhow::Result<crate::tls::TlsConnector> {
...
+    }
+}
+
+fn h2_round_error(error: h2::Error, context: &'static str) -> RoundError {
+    let transport =
+        error.is_io() || (error.is_go_away() && error.reason() == Some(h2::Reason::NO_ERROR));
+    let error = anyhow::Error::new(error).context(context);
+    if transport {
+        RoundError::Transport(error)
+    } else {
```
Zakkaus, line 567: A remote `REFUSED_STREAM` lands in `Invalid` here and fails the measurement at line 669, where `origin/main` fell back to the warm sample. RFC 9113 §8.7 says a `REFUSED_STREAM` request was not processed and is safe to retry, so it is not evidence of a malformed response. Something like

```rust
let refused = error.is_reset()
    && error.is_remote()
    && error.reason() == Some(h2::Reason::REFUSED_STREAM);
let transport = error.is_io()
    || (error.is_go_away() && error.reason() == Some(h2::Reason::NO_ERROR))
    || refused;
```

keeps the local refusal that `h2` generates for an oversized header list (`is_remote()` false) on the `Invalid` side.

Zakkaus (commented):
Read the diff against `27426b2c` and ran the gates locally on `d4e6b3a1`: fmt clean, clippy clean, 2286 tests pass / 0 fail / 4 ignored.

One behaviour regression I think should be fixed before merge, inline below: `h2_round_error` only treats I/O errors and `GOAWAY(NO_ERROR)` as transport failures, so a remote `RST_STREAM(REFUSED_STREAM)` on the measured request now fails the whole measurement. On `origin/main` any second-round error fell back to the warm sample (`urltest.rs:568`), so a server that refuses one stream (concurrency limit, graceful shutdown) used to yield a valid RTT and now yields 503 from `/proxies/{name}/delay`; repeated, that demotes the node.

Three behaviour changes the PR body and docs do not mention, in case they are unintended:

- `urltest_node_addr` now maps a zero timeout to the 15 s default (`urltest_timeout`); `origin/main` passed zero through.
- Score evidence for HTTP/1 is now first-byte timing rather than the time of the full warm head, so the recorded values differ from before even when the measurement is unchanged.
- The User-Agent is `honk-http-probe/1.0` for both health and URLTest; it was `honk-health/1.0` and `honk-urltest/1.0`. Anyone filtering check traffic server-side by UA will see the change.

Non-blocking, take or leave:

- The target is decoded and validated several times per probe (decode the URL, serialise a request, decode it again for resolution, normalise again in measurement). Keeping the canonical target from the first decode would remove most of that.
- `probers.rs` reintroduces default-port and bracket handling for Score attribution that `honk_config::check` already owns.
- The new tests do not exercise a successful HTTPS→H2 exchange through `probe_http`, an H2 header-limit rejection, or the remote-refusal case above; the H2 tests call the private exchange helpers directly.
- `doc/zh/design/groups.md:143-145` and `doc/zh/reference/api.md:58` read as translated implementation notes ("保留可失败的节点准入", "消费临时响应头"); "先校验节点，校验失败时不创建资源" and "先读取并跳过信息性响应" say the same thing in plain Chinese.

Glassyiris:
Tested head `d4e6b3a1f123be150bb685226b310a42f377df14` on the Debian 13 x86_64 test host (Linux `6.12.94+deb13-amd64`). This was a manual static-musl build, not a release artifact: an isolated `--mock-ebpf` process with private ports/data, real outbound networking, and temporary loopback HTTP/SOCKS5 fixtures. The existing real-datapath service was not replaced or restarted.

- All 12 HTTP/API scenarios passed: raw path/query and credential/fragment handling, IPv6 authority, 103 informational responses, validated first-response close fallback, first/second 503 rejection, partial-head timeout, oversized head, HTTPS-to-plaintext rejection, group delay, and periodic HEAD→configured GET.
- Three consecutive periodic 503 responses produced no new successful TCP latency sample; healthy samples resumed after recovery.
- 33/33 actual Hysteria2→HTTPS API probes succeeded, with 33 negotiated `h2` observations. In the 20-call repetition, measured RTT was 47–53 ms (median 50 ms); no retained Hysteria2 warm session was reported.
- `honk-tool sub` also succeeded: v4 51 ms, v6 51 ms, URLTest 49 ms, UDP DNS 49 ms, QUIC handshake 186 ms (single run).
- Resource nuance: a three-second FD check was too early. A follow-up 10-call burst added 10 UDP descriptors while QUIC connections were closing; the process returned to its 23-FD baseline about 10 seconds after the final probe. This is bounded observed drain behavior, not a claim of immediate socket release or a long-duration leak proof.

Both temporary services exited with code 0. Their processes/listeners and private test directory were removed; the original service stayed active with the same PID, and its config, executable and BPF-object SHA256 hashes remained unchanged. No repository source changes were needed. The explicitly deferred upstream limitation https://github.com/hyperium/h2/issues/958 remains open. This does not count as real eBPF/NFQUEUE datapath or kernel-matrix validation. The temporary harness was LLM-assisted and actually executed on the target host.

PR #246, opened 2026-09-14
feat(vless)!: align UDP multiplexing and harden outbound boundaries

Glassyiris:
## Problem

VLESS links could remain [UDP-incapable Legacy](https://github.com/daeuniverse/honk/blob/c2cec879d072d7fe543391e7fcbf3f295de2cd41/crates/honk-outbound/src/descriptor.rs#L62-L64) despite `udp=1`, conflating permission, framing and multiplexing. The review also exposed premature source retirement, cached typed probe refusals, capacity/health misclassification, feature-off descriptor reservation, and discarded or conflicting TLS assertions.

## How I fixed it

Separate canonical UDP permission, encoding and mux policy, allowing UDP443 by default while preserving explicit policy and routing blocks. Keep carriers generation-owned and XUDP sources core-owned; retain pending attachments, distinguish missing capability from terminal policy, and preserve retryable typed probe refusals and health-neutral capacity errors. Map Trojan REALITY through the existing authenticated transport, distinguish plaintext VLESS identity, and reject conflicting TLS declarations. Retain the isolated feature-off CI coverage and downstream-only Vision limitation. This is a breaking upgrade: every VLESS ID is re-derived and `vless_mode` is removed; migrate static/provider/cache inputs first. See the [design and limitations](https://github.com/Glassyiris/honk/blob/11e353c73e82f0f9feccef795463c87f0189082e/doc/en/design/outbound.md). These changes are LLM-authored with Oh My Pi under the maintainer's section 11 authorization; human approval remains required.

## Verified

Head `11e353c` passed workspace tests (**2506 passed, 7 ignored**), both outbound feature gates, workspace/all-target Clippy, feature-off core budget tests and toolbox tests, and feature-off core/tool Clippy. Production-hunk ablation made the targeted regressions fail before restoration; the claimed H2 cancellation leak was not reproduced, and its change only removes unnecessary Drop locking. Earlier official-peer and deployment checks retain their recorded revisions; [published datapath measurements](https://github.com/Glassyiris/honk/blob/6ec422e265d49d44ad4577c2190e90d6891335e0/bench/results/pr246-datapath-20260916/report.json) compare `c2cec879` with `d31f024`, not this head, and do not establish throughput parity. No new live deployment or ignored public-peer/native-kernel/VM run was performed for these review fixes.


Zakkaus:
我看看 分配一下

Zakkaus:
Read at `a9619d30` (merge base `c2cec879`). Local on that head: `cargo fmt --check`, `cargo clippy --all --all-targets -D warnings`, `cargo test --workspace` (2471 passed, 0 failed, 5 ignored); release `honk-core` 19,819,216 bytes vs `main` 19,526,864 (+285 KB, the comment's +0.3 MB); DNS smoke peak 22.0–22.3 MB over three runs, the same band as `main`. Production diff is +7008/−5993, tests +9183/−541. A second read by another reviewer covered the DNS and health consumers; everything below I re-traced myself.

**1. Two DNS consumers still turn a terminal packet refusal into an answer, against the contract this PR writes down.** `doc/en/design/outbound.md:406-410`: "DNS, health, and CLI callers preserve the rejection rather than selecting a fallback." `upstream_pool/query.rs:394-396` and `service/name_resolution.rs:53-65` do that. Two outer paths were not updated: `dns/engine/pipeline/operation.rs:35-49` serves a stale cached answer for *every* exchange error when the entry is reuse-eligible, a `PacketRejection` included; `dns/forwarder.rs:606-614` treats any sibling error in the preferred-family probe as "no answers" and returns the other family's response. Both need the `is_packet_rejection` check before the fallback decision. The existing stale test (`dns/engine/tests.rs:454`) uses an ordinary offline error; the regression to add is the same expired-cache scenario with a typed Capacity refusal, plus a preferred-family refusal beside a successful non-preferred answer.

**2. Every VLESS node changes ID at this upgrade, not only nodes that carried `vless_mode`.** The dial shape now includes the effective UDP permission, fallback encoding and multiplex paths, so a plain `vless://…?security=tls` link derives a different ID on this branch than on `main` (`95b2eabc…` → `fc8ff7f9…` for the same link; Vision+REALITY likewise). `doc/en/reference/nodes.md:42` says VLESS nodes "may also receive a new ID when migrated from `vless_mode`"; the PR body says nothing. What survives: selector choices (by member name) and persisted delay samples younger than 24 h (`control/cache.rs:42-74` re-keys them by name onto the new ID and seeds ranking, not liveness). What resets: ID-keyed health and warm state. Say "every VLESS node" and list those two exceptions, so nobody deletes selector state for nothing.

**3. `vless_mode` in `node {}` is a fatal configuration error; in a subscription it drops one node; everywhere else in the parser an old spelling gets a `legacy-*` diagnostic and keeps working for a release.** `parser/entries.rs:112-114` routes `removed-vless-mode` to `ParseFailure` next to unknown-protocol, so a config that worked yesterday refuses to load (`Parse error: nodes[1].vless_mode: VLESS vless_mode was removed; use packetEncoding and mux`), while `subscription.rs:684-698` demotes the same error to a per-entry warning. The migration table in `nodes.md:283-297` is exactly a mapping the parser could apply under a `legacy-vless-mode` warning, which is how #233 handled every other renamed or re-read field. If hard rejection is the intent, that is a maintainer decision; then make the two paths behave the same, mark the commit as breaking per Conventional Commits (`feat(vless)!:` or a `BREAKING CHANGE: vless_mode removed; see doc/en/reference/nodes.md` footer), and tell operators to edit static links and provider bodies *before* upgrading — a stored subscription body that contains only old-mode VLESS entries restores nothing offline.

**4. The new rejection classes reach the user without their reason.** `apply_vless` (`share_link/options.rs:533-560`) rejects about twenty third-party spellings (`packet-encoding`, `smux`, `xudp`, …) with a precise message, and the composition rules reject e.g. Vision with `mux=h2mux`; both surface as the pre-existing `invalid-node-entry` / "node entry could not be parsed; ignored" (`entries.rs:116-121`), the reason discarded. Before this PR that path caught malformed links; now it catches valid links written the way sing-box or Clash Meta write them. Carrying `error.diagnostic.message` into the notice keeps the redaction rule (the messages name parameters, not values) and makes the drop actionable.

**5. Docs contradict the implemented Vision UDP/443 exception.** `doc/en/design/outbound.md:491-493` (and zh `:432-434`) says UDP/443 is admitted only by `allow` on an actual UDP mux path; `nodes.md:199` says a nonempty flow "must be exactly `xtls-rprx-vision`". The code (`node/vless.rs:420`, `:269`) blocks 443 only for the base flow, and `flow=xtls-rprx-vision-udp443&mux=off` sends port 443 over Single XUDP — which is what the flow table at `nodes.md:306,333` says. Fix the two paragraphs, not the code.

**6. The carrier budget is carved out of the descriptor budget unconditionally.** `resource_budget.rs:58-61` reserves `after_dials / 8` for VLESS carriers before sizing UDP endpoints, whether or not any node uses `mux=`. From the checked-in formula: `nofile` 1024 → 50 to 44 UDP endpoints, 4096 → 216 to 189, 65536 → 4588 to 4015 (12 % fewer in every common case; only the 1,048,576 cap is unaffected), and exhaustion is an immediate typed Capacity refusal, not a queue. The body says "can reduce UDP endpoint capacity"; the number belongs in `doc/en/design/outbound.md` next to the partition, or the reserve should be zero when the loaded configuration has no multiplexed VLESS node (the budget is recomputed on reload anyway).

Notes, no action needed from me:

- The bundled fixes read correct and are separable: a bare pooled socket is now usable only when a non-blocking `MSG_PEEK` returns `EWOULDBLOCK` (`pool.rs`), so a server that already sent an alert never reaches a handshake; REALITY offers `X25519MLKEM768:X25519` with Chrome-style shares and rejects a repeated ClientHello without resealing (`reality.rs`, `reality/wire_tests.rs`). An older REALITY server that does not know the hybrid group selects the X25519 share, which the transcript test covers. Two commits would have made them reviewable on their own.
- `packetEncoding=auto` matches Xray v26.9.9 `proxy/vless/outbound/outbound.go` under its default `cone` setting (Vision, or `cone && port ∉ {53, 443}` → XUDP; otherwise native), not every Xray environment.
- Two of the new tests assert private accessors rather than behaviour (`runtime/tests/vless_runtime.rs:4-72`, `quic/packet_transport.rs:543-550`); the wire tests beside them (`proxy/vless/tests/pools.rs`, `udp_endpoint/source_tests`) are the ones that carry the proof.
- The PR body's test and interop numbers are for `2b8e891e`, not the merge head; the older interop fixture still dials `www.cloudflare.com:443` for REALITY and takes `sing-box` from `PATH`.

Not checked: the XUDP source/session ownership state machine (`udp_endpoint/source*`, about 2,500 lines) beyond its tests passing; any live peer; the recent-kernel lane.

Zakkaus:
补一份按质量判据的整体评价，位置都对应上面那条留言里的发现。

| 判据 | 评价 |
|---|---|
| 逻辑 | 核心模型成立：三个独立选项、`normalize` 收敛到规范形、`auto` 与 Xray 默认一致，终止性拒绝的契约定义清楚。边缘没有收完：DNS 两处 fallback 漏了，文档与代码在 `xtls-rprx-vision-udp443` 上不一致。 |
| 可维护 | 九千行测试，wire 测试看的是行为，这是正面。负面是 `VlessConfig::normalize` 三十行互相牵制的规则（Vision、udp443、Protocol 路径）没有一张表说明，下次修改容易踩坏；一个 17k 行的提交无法 bisect，三个不相关的修复混在其中。 |
| 可读性 | 命名一致（`VlessTcpPath`、`VlessUdpPath`），中英文档同步更新；文档两处自相矛盾，`normalize` 缺注释。 |
| 通用 | 分路径的池、进程级 FD 预算，形状是通用的；预算切分是静态的，没有 mux 节点也切走 12 %。 |
| 易用 | 弱项：`vless_mode` 硬拒载、二十种拒绝不给原因、迁移表只写在文档里而没有给解析器使用。 |
| 兼容 | 破坏性变更：删除配置键、所有 VLESS 节点 ID 重算，提交没有标 `!` 或 `BREAKING CHANGE`；第三方写法直接拒绝。好的一面是默认行为对齐了 Xray。 |
| 其他范围的影响 | 波及面大：DNS、健康探测、URLTest、Score、tool、bench。正文列得诚实，DNS 那两处就是波及没有收干净的证据。 |
| 不过度设计 | 大体没有。多出来的机制是两个只断言私有访问器的测试，以及无条件的 FD 预留。 |
| 不过度修复 | REALITY key share 与 bare 池排队字节两项修复与本次特性无关，应各自一个提交。 |
| 不偷懒 | 相反，做得很足；只有「may receive a new ID」那句轻描淡写。 |
| 面条代码 | 读过的部分不是。`udp_endpoint/source` 约 2,500 行的状态机没有逐行读，不下结论。 |

总评：设计中心站得住，测试是认真的。质量问题出在两处。一是边缘没有收完。二是 PR 的形状：一个大提交、混入了修复、破坏性变更没有标记。

Glassyiris:
Addressed in 788599f, c968a66 and a96cbf7. Current head: `a96cbf7034cacd3a4a575db2fc1e40f5c9cf3736`.

1. Confirmed both DNS bugs with the actual forwarder: Capacity became a stale answer, and a refused preferred-family lookup returned the successful non-preferred answer. Both outer fallback guards now preserve the original typed cause. Regression tests cover strict/compatibility paths, both family preferences, wrapped errors, and ordinary offline fallback.
2. EN/ZH upgrade notices now state that every admitted VLESS identity changes, including omitted-mode and ALPN-bearing nodes. They explain conditional name-based Selector restoration and persisted TCP-v4 delay samples no older than 24 hours; those samples do not restore liveness.
3. The previously selected clean break remains: no `vless_mode` compatibility alias. Static configuration failure and subscription per-entry salvage are intentional ownership boundaries, now documented together with offline all-old-mode cache recovery. The PR title marks the break; published history was not rewritten.
4. Fixed both diagnostic-loss layers: exact allowlisted VLESS errors now become static field/reason diagnostics, then static/provider admission retains their safe metadata. Unknown prose, raw values, links and credentials remain redacted.
5. Corrected both Vision paragraphs: the `-udp443` suffix permits protocol-fallback Single XUDP at port 443 even with mux off; enabled Xray reject/skip/allow precedence remains unchanged. Added brief normalization invariants rather than restating each branch.
6. Published the 50→44, 216→189 and 4588→4015 capacity figures. The budget is fixed at startup, not recomputed on reload, and Native/UoT/Single-XUDP also consume carrier permits; setting it to zero merely because no mux pool exists would break those paths. No dynamic repartitioning was introduced.

Verification: config 467 passed; workspace 2457 passed / 7 ignored; dns-ci, formatting, workspace/all-target Clippy and no-default-feature core Clippy passed. Five focused negative controls failed with the corresponding production fix removed, then passed after restoration. Two independent scoped reviews completed hostile rechecks. Wire implementations were unchanged in this follow-up, so the earlier remote interop evidence is explicitly attributed to its earlier revision rather than presented as a new run.

Zakkaus:
Re-read the fixes on the current head. All six are addressed as described: the two DNS fallbacks now keep a typed refusal (`dns/forwarder.rs:609`, `dns/engine/pipeline/operation.rs:37`), the upgrade notice says every re-derived VLESS node changes ID and names the two things that survive by name, allowlisted VLESS share-link errors reach the user with their own code instead of the generic drop (`parser/entries.rs:112-120`), the Vision `-udp443` paragraphs match the code, the capacity table carries the 50→44 / 216→189 / 4,588→4,015 figures and the reason the reserve stays fixed, and the title marks the break. Nothing further from me on this PR.

Glassyiris:
## Paired real-datapath benchmark results

Compared PR merge base **`c2cec879`** with candidate **`d31f024`**. The later `901f563` commit changes tests only; its production code is identical to the measured candidate. Evidence is pinned in [`bench` commit `6ec422e`](https://github.com/Glassyiris/honk/tree/6ec422e265d49d44ad4577c2190e90d6891335e0/bench/results/pr246-datapath-20260916), including raw attempts, engine logs, binary/config hashes, fixture inputs and restoration checks.

### Method

- Engine hosts: x86 `10.10.10.49` / Linux 7.2.3 and ARM64 `10.10.10.118` / Linux 6.12.94. Client traffic crossed the real eBPF/NFQUEUE/TPROXY datapath from an owned network namespace.
- Release musl builds with `mimalloc`, `rprx` and `ebpf`; two Tokio workers. Official peers: Xray 26.9.9 and sing-box 1.13.19 on `10.10.10.70`.
- Three alternating baseline/candidate pairs per host. Each case per arm/run performed one HTTP warmup, 15 HTTP opens, three 16-MiB HTTP downloads, three 16-MiB HTTPS downloads, and 100 sequential 1200-byte UDP echoes to each of two destinations using one client socket.
- Exact body hashes, UDP nonce/sequence/payload, reply addresses and independently observed upstream sender identity were checked. Baseline legacy settings and candidate canonical settings selected equivalent paths at the measured ports. Failed attempts were retained.

### Results

Values below are **baseline → candidate** medians. Download rates are Mbps; UDP echo latency is ms. These finite-transfer rates include connection setup and client hashing, not pure NIC or engine throughput.

| Host | Path | HTTP download, Mbps | HTTPS download, Mbps | UDP echo, ms |
|---|---|---:|---:|---:|
| x86 | Direct control | 7136.9 → 8128.9 | 6446.3 → 6121.8 | 0.070 → 0.067 |
| x86 | Single XUDP | 7140.8 → 7176.0 | 4224.2 → 4154.6 | 0.210 → 0.190 |
| x86 | H2MUX | 4925.6 → 5110.8 | 5488.4 → 4710.6 | 0.216 → 0.220 |
| x86 | Vision + XUDP | 5218.3 → 4804.8 | 2070.8 → 5073.4 | 0.273 → 0.212 |
| x86 | REALITY + XUDP | failed → 4846.4 | failed → 4668.8 | failed → 0.224 |
| ARM64 | Direct control | 363.2 → 390.7 | 216.2 → 222.5 | 1.629 → 1.690 |
| ARM64 | Single XUDP | 327.1 → 288.4 | 171.0 → 167.8 | 2.747 → 2.729 |
| ARM64 | H2MUX | 304.6 → 298.0 | 160.0 → 161.0 | 8.282 → 8.236 |
| ARM64 | Vision + XUDP | 147.6 → 147.3 | 159.8 → 160.6 | 3.097 → 3.131 |
| ARM64 | REALITY + XUDP | failed → 146.8 | failed → 155.2 | failed → 3.174 |

**Candidate: 3330/3330 attempts passed on each host, 6660 total with zero failures.** Both hosts completed all 30 case groups. Baseline REALITY failed all 666 attempts on each host (1332 total); all other baseline groups passed. The complete A/B drivers therefore correctly exited nonzero—not because the candidate crashed, and not an all-green baseline/candidate result. No baseline REALITY throughput comparison is available.

### Interpretation and limits

Performance is **not uniformly improved**: x86 H2MUX HTTPS median decreased **14.2%**, ARM64 XUDP HTTP decreased **11.8%**, and x86 Vision HTTPS increased **145.0%**. Direct controls also varied. No controlled ablation isolates the causes, so this evidence does **not** establish throughput parity or zero performance regression.

This matrix covers four VLESS proxy paths plus direct controls on IPv4 LAN. It does not establish native-UDP coverage, other-protocol performance, full Vision uplink shaping, long-duration stability, privacy guarantees or the ignored kernel/VM test suites. HTTPS used an explicitly unverified test certificate. Superseded-baseline and ARM setup failures are archived separately and excluded from these measurements.

The ARM run used held-namespace-FD entry and a temporary `fw4` rule restricted to the owned client interface and four direct-control ports. Original service, binary/config/cron hashes and the authenticated API were verified restored; owned interfaces, namespace, nft table and forwarding rule were removed. All owned upstream fixtures were stopped.

[Machine-readable comparison and conclusions](https://github.com/Glassyiris/honk/blob/6ec422e265d49d44ad4577c2190e90d6891335e0/bench/results/pr246-datapath-20260916/report.json)

Glassyiris:
LGTM， soaking in vyos

Zakkaus:
Third pass, on the 29 commits `a96cbf70..901f5631`. Gate on the head: `cargo fmt --check` clean, clippy clean, workspace nextest 2481 passed / 0 failed / 7 ignored. All six findings from the previous round are fixed as described; nothing blocking in the delta. Three things worth doing before merge, one of them measured.

### 1. `6b8037ef` shrinks the relay read from 64 KiB to 8 KiB, and that is visible on the wire

The flush before the copy loop is right and the new duplex tests pin it. But the same hunk also replaces `BufReader::with_capacity(64 KiB)` + `copy_buf` with plain `tokio::io::copy`, whose internal buffer is `DEFAULT_BUF_SIZE = 8 * 1024` (tokio 1.53.1, `src/io/util/mod.rs:88`). Every framed outbound now gets 8 KiB writes instead of 64 KiB ones: for AnyTLS that is eight frames, eight `Bytes` allocations, eight permits and eight headers where there was one (well, 65,535 + 1, see #253), and H2/gRPC bodies are now chunked at 8 KiB instead of at the frame maximum. This is the most plausible source of the H2MUX HTTPS −14 % in the benchmark comment, which so far has no attribution.

Measured on the AnyTLS lab pair from #253 (same method, two alternating rounds, `iperf3` uploads through the eBPF datapath, release build of `901f5631` vs `main` `19d8125`):

| build | relay read | p1 (Gbps) | p4 (Gbps) | reverse (Gbps) | p4 CPU s/GB | peak RSS (MB) |
| --- | --- | ---: | ---: | ---: | ---: | ---: |
| `901f5631` (this PR) | 8 KiB | 3.8 / 3.7 | 9.0 / 8.9 | 5.3 / 4.3 | 0.63 | 63 / 70 |
| `main` `19d8125` | 64 KiB | 6.1 / 9.4 | 17.0 / 17.4 | 5.9 / 5.9 | 0.59 | 128 / 127 |
| #254 (`main` + 65,535-byte read + byte-bounded queue) | 65,535 | 6.3 / 13.1 | 21.7 / 21.2 | 7.5 / 5.9 | 0.63 | 88 / 75 |

Two values are the two rounds; UDP and latency are unchanged across the three (~0.7 Gbps, 0.05 ms). Upload throughput halves at both one and four streams while CPU per GB stays where it was, so this is the per-frame cost in the serialised session writer, not compute. Peak RSS drops only because the frame-counted queue now holds 8 KiB frames instead of 64 KiB ones, which is the (b) point of #253 by accident rather than by design.

Suggested change: keep `wr.flush().await?` and the tests, restore the sized reader:

```rust
wr.flush().await?;
let mut br = tokio::io::BufReader::with_capacity(
    RELAY_BUF_SIZE,
    ReadCounter::wrap(rd, progress, None),
);
let n = tokio::io::copy_buf(&mut br, wr).await?;
```

Note this hunk also conflicts textually with #254, which makes the same constant `u16::MAX` so a saturated read fits one AnyTLS frame; whichever lands second needs a one-line rebase.

### 2. Backend carrier closure does not wake pool capacity waiters

The delta fixes the release, publication, warm-unpin and janitor wakeups (`notify_waiters` + `Notified::enable()` is the right pattern), but `VlessMuxSession::driver_finished()` (`crates/honk-outbound/src/proxy/vless/mux.rs:98`) stores `Closed` and closes its own stream semaphore without notifying the owning pool, and a waiter parked in `Step::Capacity` (`session.rs:450`, `:521`) listens on `capacity_notify`, not on the backend semaphore. `reap_unretained_idle()` (`session/maintenance.rs:33`, `:57`, used by `runtime/vless.rs:148`) drops an already-closed entry without adding it to `to_close`, so it notifies nothing either. Sequence, not reproduced: two carriers full, a checkout parked, one carrier's driver ends while callers still hold its child streams (and their `SessionPermit`s) — a replacement dial is now admissible but the waiter stays asleep until an unrelated release or its own deadline. Pre-existing rather than introduced here, but it is the one remaining hole in the wakeup contract this PR set out to close. Fix is a notify on backend terminal transitions plus on capacity-releasing manual pruning, with a test that parks a waiter, closes a carrier and requires a replacement dial before the child permits drop.

### 3. The new feature-off tests never run in CI

`crates/honk-outbound/tests/feature_boundary.rs` and the `honk-tool` sub test are gated on `not(feature = "rprx")`. The workspace nextest lane enables `rprx` through members' default features, so both are compiled out; the independent-features lane only runs `cargo check` and is push/full-CI only; `ci/outbound-ci.sh` has the isolated feature-off invocations but no workflow calls it. They pass when run in isolation (`cargo nextest run -p honk-outbound --no-default-features --test feature_boundary`), they just are not part of any gate. Add the isolated invocations to the lane that owns feature combinations.

Smaller notes, no action needed unless you want them: `a9da013d` is characterisation of downstream-only Vision, not a regression test (it passes on the previous production code too), and the doc now says so; the weekly public-VLESS lane adds scheduled runs for tests that were already ignored, not a change to the PR gate. The PR body matches the current scope.

Glassyiris:
Addressed all three items in `4183cba`, `c3144f8` and `0468375` (head `046837554e11dfcf7d67ec8b85616a1d92070c59`).

1. Restored frame-sized directional reads at **65,535 bytes**, matching #254's frame boundary without taking its byte-queue changes. The initial flush remains, along with flushing on idle input and EOF and the existing half-close/drain behavior. I did not use the suggested `BufReader + copy_buf` verbatim: restoring it made `relay_tcp_flushes_buffered_request_without_client_eof` time out again, because Tokio's `copy_buf` returns on a pending read without flushing. The small sized-copy loop preserves that contract. The saturated/partial-write regression observes 24 frames with the 8-KiB control versus three with the fix, with identical payload/accounting.
2. Bound each production session backend to its owning pool's existing capacity notification before publication or detached attachment. H2 driver termination, backend close/drain and manual removal of already-closed carriers now wake waiters. The real-H2 regression holds both carriers' 128 child permits, terminates a peer and requires speculative admission plus a replacement dial before releasing those permits; it covers ordinary, detached and inserted publication. Removing the terminal wake makes it time out. Removing the closed-prune wake independently makes the manual-pruning regression time out.
3. Added isolated outbound `feature_boundary` and full toolbox feature-off test invocations to the existing Independent feature sets lane. Each asserts the exact regression appears in the test list before running; an empty-list negative control fails rather than reporting green. Existing lane triggers are unchanged.

Verification: seven relay tests, both wakeup regressions and the exact new CI commands passed; toolbox feature-off suite 26 passed. Both outbound feature gates, workspace tests (**2484 passed / 7 ignored**), workspace/all-target Clippy, feature-off core/tool Clippy, offline documentation links and the two pinned official public loopback cases passed. Independent scoped hostile reviews found no remaining defect in these fixes.

No new throughput-parity claim: the published performance numbers are still for `d31f024`, not these commits, and the current one-hour VyOS soak deliberately remains on `901f563` so its evidence is not mixed with this patch. The AnyTLS byte-queue change in #254 remains separate.

Glassyiris:
The VyOS soak exposed a separate real UDP routing defect, fixed in `856408bea9e92a0618bcea86970954b4c715e0b9`.

A selected nested Score/Selector group could have no UDP-healthy member and an explicit healthy `final`, yet recursive selection ignored that child final. The outer requested group had no own final, so it produced `No available candidate nodes for UDP outbound` and dropped the flow. The configured Hysteria2 backup passed DNS/QUIC checks; the two sampled VLESS paths timed out UDP DNS in both honk and an official sing-box control, so this was not repaired by bypassing health or guessing a wire-format change.

GroupManager now resolves explicit final node/group/direct/block paths at each nesting level, preserving Selector authority, Peek/Apply, bounded cycles/depth, IPv4 proxy-health preference for IPv6 targets, selection chains and Score attribution. Health registration, warm discovery, Score membership and datapath reachability include final-only leaves; health changes update every affected ancestor slot rather than only the first mapped group. Named finals retain the first configuration declaration when node names collide. The duplicate core final resolver was removed. No gateway routing/configuration edits or warning suppression were used for this fix.

Verification: 11 new behavioral regressions fail under removal of the corresponding fallback/health/attribution/identity changes, then pass after restoration. Both outbound feature gates, workspace tests (**2495 passed / 7 ignored**), workspace/all-target and feature-off core/tool Clippy, and documentation checks passed. Two scoped hostile reviews found and then verified the duplicate-name correction. On the gateway, the same LAN HTTP/3-only request changed from a five-second timeout with a matching empty-plan warning to **HTTP 204 / HTTP/3** on both IPv4 (~178 ms) and IPv6 (~200 ms); connection tracking shows the configured nested-final chain ending at Hysteria2, not direct. Configuration bytes were unchanged in this deployment. The corrected version's one-hour process and dual-stack HTTP/3 observations are running; they are not yet claimed complete.

Correction to the initial operational audit: journald represented ANSI-containing messages as byte arrays. My initial string-only scan missed warnings; after decoding, the original `901f563` hour contained 2095 warning lines. Its 241 process samples and 61 DNS/HTTPS probes passed, but that did **not** establish UDP health and is not an all-protocol success claim. The full raw gateway logs remain private because they contain traffic metadata.

olicesx:
基于 `901f563` 审阅。前 9 条是本 PR 引入的；后 3 条是既有问题（在 base `c2cec879` 上用相同用例验证过行为一致），但本 PR 新增了对应不变式/文档，所以一并列在这里决定。行号以 `901f563` 为准。

**1. `crates/honk-core/src/control/udp_endpoint/source.rs:205` — attachment 与 retirement 的竞态**

`release_binding` 只看 `bindings == 0` 就退休，而 `drop_attachment`（:188）要求 `bindings == 0 && attachments == 0`。一个已 attach、还停在 `driver.wait_ready().await`（`control/connection/udp.rs:699`）的 flow，其 `commit_attachment` 会被拒，`commit_ready` 随之失败，首个数据报以 "generation was cancelled before ready commit" 丢弃。只要有任一 sibling view 在这个窗口内退休就能触发。修法：:205 补上 `attachments == 0`——attachment 最终必然 committed 或 dropped，不会泄漏 source。

**2. `crates/honk-outbound/src/control/probers.rs:721-724` — 吞掉类型化拒绝**

这里把任何错误都映射成 `None` 加一条 warn，而上面的 `resolve_udp_check_target` 是保留类型化拒绝的（:670-672）。`runtime.rs:710-714` 会把这个 `None` 缓存整个 generation，`probers.rs:419-421` 于是把 `data_allowed` 置为 false，所有节点的 QUIC data-path 探测被永久关闭直到 reload——正是新写的 AGENTS.md/dns.md 契约禁止的 fail-open。请照 UDP target 的 `OnceCell` 重试方式处理。

**3. `crates/honk-outbound/src/proxy/vless/descriptor.rs:181-185` — 能力拒绝被当成终止性 policy**

`udp_target_allowed` 经 `udp_path`，而 `udp_path` 在 `!udp_enabled()` 时返回 `None`，于是 VLESS 在 `registry.rs:339-343` 先撞到 `PacketRejection::Policy`，`udp_capable` 对 VLESS 已成死分支。ColdUrlTest 里这会在 `udp_dial.rs:90-93` 直接 break：sibling 不启动、`on_dial_error` 也不执行。Trojan/AnyTLS 同样带 `network` 门时走的是普通错误。是刻意如此吗？是的话在调用点写清楚；不是的话交给 capability 路径处理。

**4. `crates/honk-tool/Cargo.toml:11-13` — rprx-off 模式没有任何 CI 编译**

`default = ["rprx"]` 加上新加的 `cfg(not(feature = "rprx"))` 测试，意味着 off 模式的代码只有在有人记得传 `--no-default-features` 时才会被编译。而没有任何地方传：`ci.yml:1208` 是普通 `cargo check -p honk-tool`，也没有 workflow 调用 `ci/outbound-ci.sh`（那里只覆盖 `honk-outbound`）。要么加 lane，要么删掉这个测试。

**5. `crates/honk-core/src/control/resource_budget.rs:58-61` — 无 rprx 时的 carrier 预留**

`vless_carriers` 被无条件扣减，但 `rprx` 关闭时 `runtime.rs:939-945` 直接丢弃该上限、信号量根本不存在。feature-off 构建白白少了 UDP endpoint（`nofile 1024` 下 44 vs 50），而新写的文档句子说的是相反的话。请用同一个 cfg 门控这个扣减。

**6. `doc/en/design/outbound.md:105-106` — 过时的判定函数**

文档说 `network_allows_udp` 是共享的能力判定，但 VLESS 实际调的是 `VlessConfig::udp_enabled()`（`descriptor.rs:62-64`）。对已准入节点行为一致，但这句话现在会把维护者指向一个孤儿 helper。要么改用该 helper，要么删掉这句。

**7. `crates/honk-outbound/src/session.rs:721-729`（另见 `session/speculative.rs:161-168`）— 未类型化的容量错误**

stream 槽耗尽返回裸 `anyhow!("session has no stream capacity")`，与 `runtime.rs:648-653` 的类型化 `PacketRejection::Capacity` 不一致。正常饱和会 park（health 中性），但 post-`Have` 抢最后一个槽的 stampede 会落到这里，被计成节点故障。若非有意，请补类型。

**8. `crates/honk-outbound/src/proxy/vless/mux.rs:847-852` — `Drop` 里的 `try_lock`**

writer 锁被占用时 `Drop` 会跳过 `send_reset`，于是被取消的在途 UDP send 所占的 H2 stream 会一直挂到 carrier 死亡。`MuxSendStream::drop`（:528-531）是无条件 reset 的。两者应统一。

**9. `crates/honk-core/src/control/udp_endpoint/source_tests/regressions.rs:88` — 新测试依赖真实时钟**

`start_paused = true` 配真实 loopback peer：2s 的 commit deadline 可能被 tokio 的时间自动推进提前耗尽，于是 `:51` 以 "connect timeout" 失败。本机 3/3 失败，CI 却是绿的。建议改用真实时钟或确定性信号，而不是这个 deadline。

---

以下不是本 PR 引入的，但本 PR 新增了对应不变式、也动了这个解析器，建议在这里一并决定：

**10. `crates/honk-config/src/share_link/options.rs:281-291` — Trojan REALITY 被静默降级**

`trojan://pw@host:443?security=reality&pbk=AAA&sid=ab&spx=/&sni=mask.example` 会被接受，但 `pbk/sid/spx` 从不拷贝，`effective_reality_public_key()` 得到 `Ok(None)`，最终是拿普通 PKI TLS 去连 REALITY 服务器。同样 query 的 VLESS 是保留的。要么对不使用这些 key 的 scheme 拒绝 `security=reality`，要么给 Trojan 补上映射。

**11. `crates/honk-config/src/node.rs:264` — `derive_id` 不含 `tls.enabled`**

`vless://UUID@host:443?security=none` 与 `vless://UUID@host:443` 派生出同一个 UUID（`fd9d120c-…`），`Config::validate` 随即以重复 ID 拒绝。也就是说刷新时把 TLS 切成明文，ID 不变，Score/health 历史会跨安全姿态变更被错误沿用。base 同样如此，但本 PR 重算了所有 VLESS ID，正好一并纳入这个标志。

**12. `crates/honk-config/src/share_link/options.rs:159-190` — 重复的 `security`/`tls` key 后者胜出**

`?security=tls&security=none` 被接受并静默变成明文；反过来则启用 TLS。其它重复的 VLESS 控制项都会被 coalesce 或拒绝，这两项应对齐。

---

另：PR body 里的 "3,105 syntax bodies / 31 deltas" 无法从仓库复现（没有把工具提交进来），reviewer 没法核验这条机械重排声明。

Glassyiris:
Addressed the latest review in `fbebc92`, `aba340e`, `19814d1`, and `11e353c` (head `11e353c73e82f0f9feccef795463c87f0189082e`), rather than applying the `901f563` line numbers blindly. The earlier user-requested UDP443 default-Allow commit `ffec627` is included in this push; explicit Reject and routing blocks remain effective.

1. Source-owner retirement now requires zero bindings and zero pending attachments. The regression retires the last bound sibling, commits the pending attachment, exchanges its first datagram, then checks final attachment cleanup. Existing stale-preparation coverage now uses actual shutdown rather than relying on the premature-retirement bug.
2. QUIC health-target resolution uses a generation-scoped fallible `OnceCell`. Typed refusal reaches the probe result without being cached as disabled; later probes retry. Resolution and QUIC dialing/handshake share a bounded attempt, and local setup is not warming evidence.
3. UDP-disabled VLESS now reaches ordinary capability refusal through registry, runtime, speculative and direct-handler paths. Cold candidates may continue to a sibling; explicit UDP443 policy remains terminal and health-neutral.
4. Already fixed in `0468375`: Independent feature sets runs the isolated outbound regression and full `honk-tool --no-default-features` suite, with nonempty exact-selection checks. No duplicate lane added.
5. Feature-off core no longer subtracts VLESS carrier slots from the UDP endpoint partition.
6. Both outbound design pages now identify VLESS's canonical `udp_enabled()` rather than claiming it shares the Trojan/AnyTLS helper.
7. Both post-offer and detached stream-slot races preserve typed `PacketRejection::Capacity` through their context, including neutral Score classification and slot rollback/reuse.
8. Removed the unnecessary `try_lock` from UDP transport Drop in favor of exclusive `get_mut`. The stronger leak claim was not reproduced: a pending send owns/borrows the transport, and its guard is released before final Drop; the blocked-send cancellation check reset the peer and reclaimed same-carrier capacity on both old and new code. This is a simplification, not a claimed reproduced leak fix; the temporary check was removed.
9. The real-loopback congestion regression now uses real Tokio time; it still holds the sender futures pending to force the queue-timeout interleaving.
10. Trojan already has authenticated REALITY in the shared transport stack. Share-link normalization now maps its REALITY assertions into that existing path; incomplete or unsupported intent rejects instead of silently becoming PKI TLS.
11. Plaintext VLESS receives distinct identity material. TLS-on/default and REALITY identities remain stable relative to the preceding branch head; unrelated protocol identities are unchanged.
12. Repeated security/TLS declarations coalesce or reject contradictions before assignment; encoded VMess cipher interpretation remains intact.

Removed the unverifiable mechanical `3105 / 31` count from the PR body instead of checking in a one-off audit tool. No repository source file crossed 1000 lines in these fixes. Independent static second passes checked the parser, probe, lifecycle and budget boundaries; this is scoped review evidence, not a blanket merge verdict.

Verification: Head `11e353c`: workspace **2506 passed / 7 ignored**, both outbound feature gates, workspace/all-target Clippy, feature-off core resource-budget tests (**5 passed**), toolbox feature-off tests (**26 passed**), and feature-off core/tool Clippy passed. Ten production-ablation invocations produced the expected behavioral failures (eleven selected tests); restoring the fixes passed the final gates. The temporary blocked-send cancellation control passed with both implementations and was removed. Touched documentation local links passed. No new deployment, live subscription probing, ignored public-peer run, or native/VM kernel run was performed in this review-fix task; earlier runtime evidence remains tied to its original revision.

PR #248, opened 2026-09-15
config: open a quote right after the entry colon

Zakkaus:
## Problem

Reported this morning: the parser does not accept `a:'b'`. It is worse than a rejection. The lexer opens a quote only at token head or right after `(` or `,`, so with no space after the colon the quote is data: `log_file:'honk.log'` stores `'honk.log'` with the quotes, `a:'socks5://127.0.0.1:1080'` in `node` drops the node with a bare "could not be parsed; ignored" warning, `domain(suffix:'example.com')` matches the suffix `'example.com'` and never fires. dae's grammar tokenizes `:` on its own, so every one of these has always been valid there, and `doc/en/reference/dialect.md` documented the gap as a rule.

## How I fixed it

`:` joins `(` and `,` as a position after which a quote opens (`crates/honk-config/src/parser/lexer.rs`). Nothing downstream changes: `split_entry`, `quoted_prefix` and `delimiter_positions` already work on the recorded spans. An unterminated quote after a bare colon is now the same located `unterminated-quote` error as after `: `; the test that pinned the old divergence (`bare_filter_quote_receives_its_own_warning`) now pins the equivalence for both spellings. Both dialect references state the rule.

## Verified

On `1e379b0d`: new lexer test (`quotes_open_after_a_colon`: single and double quotes after `:`, an apostrophe elsewhere stays literal) and parser test (`test_quoted_value_without_a_space_after_the_colon`: global, node, subscription, routing) pass; with `lexer.rs` reverted both fail. `cargo fmt --check`, `cargo clippy --all --all-targets -D warnings`, `cargo nextest run --workspace --profile ci`: 2391 passed, 5 skipped.

---


Glassyiris:
Independent local validation on `1e379b0d7e46d19b728279dc043da0a03fd24c50`, in a clean detached checkout:

- `cargo fmt --all -- --check` and `cargo clippy --offline --workspace --all-targets -- -D warnings`: passed.
- `env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY cargo test --offline --workspace --no-fail-fast`: **2,391 passed, 5 ignored**. `cargo-nextest` is not installed locally, so this was Cargo's workspace suite rather than a rerun of the nextest CI profile.
- Removed only the new `:` quote-opening predecessor locally: both new lexer/parser tests and the updated malformed-filter test failed for the expected missing-span/retained-quote/diagnostic differences. Restored the original PR source before the full gates above.
- Additional offline assertions passed for the reported `bind:'tcp+udp://[::]:53530'`, compact DNS upstream/detour, subscription URL plus exact Loon User-Agent, `name(keyword:'HK')` membership, and `domain(suffix:'example.com')`, including LF/CRLF and ordinary numeric-colon/apostrophe controls.
- Actual local DNS process smoke passed: **4/4 expected queries**, UDP, two requests on one TCP connection, unchanged SIGHUP reload, and clean SIGTERM shutdown. This used the PR's **debug binary** and a temporary copy of `ci/dns-smoke.py`, changing only the generated config to use a temporary `data_dir` and colon-adjacent quoted bind/upstream values; query traffic stayed on loopback. This is functional evidence, not a release-performance comparison.

The small shared lexer change behaves as intended in these checks. One concrete follow-up: [`.agents/rules/configuration.md:7`](https://github.com/daeuniverse/honk/blob/1e379b0d7e46d19b728279dc043da0a03fd24c50/.agents/rules/configuration.md#L7) still excludes `:` from the quote-opening rule; please synchronize it with the updated dialect references.

I also confirmed the intended compatibility change, so an explicit migration note would help: bare paths/URLs/UAs containing an unmatched literal `:'` now quote-error, while quoting the entire value preserves the bytes. For DNS, `qname(regex:'|example[.]invalid)` is now omitted with a nonterminal quote diagnostic under existing recovery, whereas `qname(regex:"'|example[.]invalid")` retains the literal regex. Quoting a raw group name is not necessarily byte-preserving because its outer quotes remain part of the name. These are consequences of the declared uniform lexical rule, not a request for per-section lexer exceptions.

No reported subscription URL was contacted and no real-eBPF/kernel tests were run. No PR source changes were retained. Validation and this feedback were produced with Oh My Pi (LLM-assisted).

Glassyiris (approved):

Glassyiris:
LGTM

PR #273, opened 2026-09-18
fix(session): stamp idle age at stream transitions instead of sampling it

### LINKED ISSUE #261: anytls: idle age is sampled by the janitor, so activity between ticks does not reset it
Split from #170. Lines are on `main` `4dd51ad1`.

A session's `idle_since` is reset only when a janitor tick observes `active_streams() > 0` (`crates/honk-outbound/src/session/maintenance.rs:188-193`). A stream that opens and closes between two ticks leaves the old timestamp in place, so a session that was busy a moment ago can be retired as idle for the whole interval before that activity.

Trigger: short streams (a DNS-over-proxy query, one HTTP request) with a janitor interval longer than the stream's life. Observed by reading the code; not exercised.

Fix direction: record the activity transition where the stream count changes (open/close), not in the sampler; the janitor then reads `idle_since` instead of inferring it.

Zakkaus:
Fixes #261.

## What changes

`ManagedSession` gains `idle_since()` and a small `IdleClock` sessions can carry. A session clears the clock in `try_reserve` when a stream slot is taken and stamps it in `permit_released` when the last slot goes back, so the two transitions the janitor never observed now record themselves. The janitor (`session/maintenance.rs`) reads `idle_since()`; a session that returns `None` (test doubles, protocols without a clock) keeps the old per-tick sampling.

AnyTLS, Mux.Cool and the VLESS mux carrier carry the clock; that is one field, one line in `try_reserve` and a `permit_released` body each.

## Behaviour table

| Input | Before | After |
| --- | --- | --- |
| Stream opens and closes between two janitor ticks | idle age keeps its old start; session can be retired one interval early | idle age restarts at the close; retirement waits the full timeout |
| Session with a stream open at a tick | not retired | not retired (unchanged) |
| Prewarmed session that never carried a stream | idle from its first zero-stream tick | idle from creation; retired one tick sooner at most |
| Session without a clock (`idle_since()` is `None`) | sampled | sampled (unchanged) |

The third row is the one deliberate change beyond the issue: a never-used session now ages from creation rather than from the first tick after insertion, which is the value the sampler was approximating.

## Test

`a_stream_between_two_janitor_ticks_still_resets_the_idle_age`: janitor every 10 s, timeout 15 s, a one-second stream at 18 s. The session survives the 30 s tick and is retired at 40 s; on the old janitor the assertion at 31 s fails because it was retired at 30 s.

`ci/outbound-ci.sh` is green. Production 92 lines, tests 66 lines.

## Checks

- Logic: the janitor's `Some`/`None` branches are reached by clocked and unclocked sessions respectively; the clock's two writers are the only stream transitions.
- Maintainability: one clock type shared by the three sessions; no second copy of the timing rule.
- Readability: comments state why the clock exists (the tick cannot see a short stream), not what the setters do.
- Generality: the fallback keeps the trait usable for sessions without a clock instead of forcing a field on every implementor.
- Usability: no new call-order requirement; sessions that already reserve through `SessionPermit` only stamp the clock.
- Compatibility: listed above; the only widened retirement is the prewarmed-session row.
- No pending review items.


Glassyiris (approved):
LGTM

PR #277, opened 2026-09-19
fix(outbound): surface gRPC and VMess rejections as stream errors

### LINKED ISSUE #274: outbound: gRPC trailer status and VMess relay-task errors are hidden as EOF
Split from #265 after #272 handled the two packet-local items. Lines are on `main` `060623fc`.

- Rejected gRPC responses can look like clean EOF: `crates/honk-outbound/src/proxy/transport.rs:576-631` ignores the response HEADERS/trailer status and treats `END_STREAM` as EOF, so a peer that refuses the stream reads as a successful dial followed by an empty read.
- VMess relay-task errors are dropped: `crates/honk-outbound/src/proxy/vmess.rs:176-218,390-397` keeps only an abort handle for the spawned relay task; a rejecting peer becomes a successful dial followed by EOF.

Both are the same shape as the closed items but need more than a scope change: the gRPC side has to read the trailer status before reporting EOF, and the VMess side has to carry the task's result back to the stream owner. Fix direction as in #265: keep the typed failure and surface it at the stream, not as EOF.

Zakkaus:
## Problem

Peer rejections in [gRPC](https://github.com/daeuniverse/honk/blob/b48f7d9cbc2ee30872823b11c9c7d7d03947b006/crates/honk-outbound/src/proxy/transport.rs#L576-L631) and [VMess](https://github.com/daeuniverse/honk/blob/b48f7d9cbc2ee30872823b11c9c7d7d03947b006/crates/honk-outbound/src/proxy/vmess.rs#L388-L398) could surface as clean EOF instead of stream errors, hiding failed proxy exchanges. Closes #274.

## How I fixed it

Move gRPC into `transport/grpc.rs` using the existing `h2` client, with 64 KiB response headers, its default 4 KiB dynamic table, and a 16 KiB owned write batch. Preserve buffered payload, terminal errors, half-close, split-task wakeups, graceful GOAWAY and IPv6 authorities; retain the VMess relay result before duplex closure. Update regression tests and bilingual docs. Upstream [h2#959](https://github.com/hyperium/h2/pull/959) is merged but not included in the locked 0.4.19. The follow-up gRPC implementation, tests and documentation were LLM-authored and reviewed with Oh My Pi under the maintainer's CONTRIBUTING.md §11 authorization; human approval remains required.

## Verified

After merging `main` at `318edf58`, `ci/outbound-ci.sh` and `cargo check -p honk-core --all-targets` passed locally. The outbound gate covers formatting, all-target Clippy and tests with `rprx` off/on, honk-config, and the 23 transport regressions; the header suite failed 9 cases on original head `47dc9ee` before the repair. The two documentation conflicts were checked against both parents, preserving the carrier-pressure and gRPC sections. Full-workspace runtime tests, real-eBPF and external-peer gates were not run locally.


Glassyiris (approved):
LGTM

PR #269, opened 2026-09-18
feat(score): verify fresh evidence and bound TCP recovery

Glassyiris:
## Problem

Score could prefer historical sample volume over current quality, let excluded candidates distort its baselines, double-count overlapping evidence, and abandon a TCP request when re-ranking still chose its failed primary. Terminal-only performance feedback obscured live degradation; pre-admission DNS timeouts could also become terminal capacity refusals, while late successful cleanup could refresh stale business-usability evidence. Fixes #267.

## How I fixed it

Separate business reliability, configured-probe quality and fresh target observations; use bounded evidence-gap validation and retractable, scope-limited verification. Publish live progress without duplicate outcomes, preserve final-edge provenance, and allow one different TCP leaf within the original deadline without replay. Classify capacity timeouts from actual pending admission, and date recent business usability from RX with expiry and failure/reload fences. Production code, tests and bilingual docs are LLM-assisted, with independent static/adversarial review; the author confirmed the CONTRIBUTING §11 exception, and human review remains required for the updated diff.

## Verified

Current-tree validation: 2587 workspace tests passed, 7 existing ignored; 117 focused Score, 8 admission, 5 core dial-scope and 6 TCP recovery tests passed. `outbound-ci` (rprx off/on), workspace all-target Clippy and fmt passed. All eight new regressions failed with their decisive production guards removed; positive gates then passed after cleaning affected package artifacts and disabling incremental compilation. Earlier userspace Trojan/Shadowsocks simulation and its limitations are documented in the [prior evidence](https://paste.gentoozh.org/XqdMxTdHdnj.md), [plan](https://paste.gentoozh.org/9UlIA9E3RDi.md) and [confirmation addendum](https://paste.gentoozh.org/4sFlMDawuAp.md); that simulation was not rerun for these fixes. No new real-eBPF/netns/VM, official external-peer or performance A/B run was performed.

PR #279, opened 2026-09-19
feat(score): preserve active incumbents and validate carrier pressure

Glassyiris:
## Problem

Follow up #269: [incumbent holding](https://github.com/daeuniverse/honk/blob/b48f7d9cbc2ee30872823b11c9c7d7d03947b006/crates/honk-outbound/src/group/score/ranking.rs#L327-L351) used historical failures and independently normalized utility, while [qualification](https://github.com/daeuniverse/honk/blob/b48f7d9cbc2ee30872823b11c9c7d7d03947b006/crates/honk-outbound/src/group/score/ranking.rs#L317-L325) depended on decaying completions, allowing recovered active paths to lose protection. Carrier congestion needs to reopen comparison without becoming a synthetic business failure. Debug builds also need one rolling release without silently retaining an old variant when a rerun has incomplete artifacts.

## How I fixed it

Retain earned qualification with genuine business RX, distinguish unresolved failure from recovered history, and compare qualified challengers against a fixed incumbent using common evidence. Feed runtime-owned TCP RTT/retransmission and QUIC RTT/loss hints into existing bounded validation, without changing reliability, backoff or live flows; preserve pooled descriptor access and avoid extra TLS boxing. Add aggregate pressure-cause counters, regressions, bilingual docs and a rolling Debug prerelease whose eight expected artifacts are checked before tag mutation. Code, tests and documentation were LLM-assisted via Oh My Pi, including independent/adversarial LLM review under the maintainer-authorized contribution exception; human maintainer approval remains required.

## Verified

At Rust head `5322af11`: 2,622 workspace tests passed (7 existing ignored), outbound-ci, workspace Clippy and fmt passed; official Xray/sing-box loopback cases passed serially after one earlier parallel connection-refused result. Real QUIC delay/loss exercises with synthetic equal qualified histories reached bounded validation without an ordinary switch. The [Debug release run](https://github.com/Glassyiris/honk/actions/runs/35438755469) passed all 10 jobs at `a897c878`; the subsequent artifact guard passed local regression, ablation, isolated bare-Git workflow checks and actionlint, not live publication. VyOS observation was not controlled A/B and proves no latency/throughput gain; real TCP impairment, repeat GitHub overwrite and root routing/netns gates were not run for these changes.


Zakkaus:
One edge in `pressure.rs::observe_carrier_pressure` (line 128): the duplicate guard is the `carrier_pressure[family]` watermark on the aggregate `Stats` cell, and that cell is LRU-evictable (`AGGREGATE_CAPACITY` 4096). `record_cell_start` recreates an evicted cell from `Stats::default()`, so the watermark is gone, while `TransportQuality::snapshot()` hands the same episode back on every heartbeat within `CARRIER_PRESSURE_TTL` (60 s). If the cell is evicted and recreated inside that minute, the same episode is accepted again: `carrierPressure` and the cause counters go up twice, and `carrier_pressure_at` reopens comparison a second time. It needs a busy enough group/node/network/family set to churn 4096 cells in a minute, so it is a large-deployment case, not a common one.

If it is worth closing, the consumed-episode watermark could live somewhere that survives cell eviction (keyed by group/network/node/family alongside `selection_reasons`, which already persists), with a test that admits an episode, evicts and recreates the cell, and rereads the same snapshot without a second increment.

Glassyiris (commented):
LGTM

PR #281, opened 2026-09-20
fix(routing): copy fact bitmaps into stack areas instead of keeping map pointers

Zakkaus:
## Problem

Native routing kept nullable map-value pointers live across later rules ([the original domain lookup](https://github.com/daeuniverse/honk/blob/d464b8eef6aadf34da5bb941d00e5f5e61f0d9c5/crates/honk-core/src/control/routing_matcher/codegen.rs#L313-L323) and [IP/MAC cache](https://github.com/daeuniverse/honk/blob/d464b8eef6aadf34da5bb941d00e5f5e61f0d9c5/crates/honk-core/src/control/routing_matcher/codegen.rs#L649-L682)), contributing to the verifier state explosion in #280. The later port-first change also left the rule-entry documentation and structural tests stale. Fixes #280.

## How I fixed it

Copy fact bitmaps into invocation-local stack areas instead of retaining map pointers. Keep domain resolution in the prologue for `domain_final`; evaluate complete positive/negative port conditions first, then resolve non-domain facts through READY guards, marking completion only after copy or zero fill. Isolate codegen tests, replace broad opcode/liveness scanning with local lookup/guard and full-port contracts, and express the unchanged 28-rule verifier fixture as dae text. Update both design docs to distinguish eager-era measurements from the lazy implementation. Retain parsed predicate hit/miss coverage as a permanent native-kernel regression using independent expected decisions. The latest test/documentation repair preserves production emission from `9365370f`; its tests, docs and independent reviews are LLM-assisted using Oh My Pi, and the lazy-emitter contribution was generated/refined with Pi agents. Human maintainer review remains required.

## Verified

[...]

Issue #280, opened 2026-09-19
bug(routing): adding two lab IP rules exceeds verifier state budget on VyOS 6.18


Glassyiris:
## Problem
On the `ui` integration commit `78bb03c35da39bdb4064768329e22b33ba173047`, adding the two narrowly scoped lab rules below to an otherwise working gateway configuration makes startup fail while publishing the initial routing policy. The verifier reaches its **1,000,000 processed-instruction limit**, not a shell argument-size limit. Because startup exits, both forwarding service and the native API/UI become unavailable.

Environment: VyOS, Linux `6.18.33-vyos`, x86_64; static musl `release-musl`, default features plus `ebpf,native-api`. Binary SHA-256: `784ec06d1b46acfcd8b25b11fb1ff3fed9a7d74f2d6779a926f789832da7704a`. Reported version: `v0.0.1.beta.80-30-g78bb03c3`. This was reproduced on that merged feature revision; it is **not** a claim of an independently reproduced failure on latest main.

## Triggering configuration
The main file includes `config.d/*.dae`. The existing `routing.dae` below worked before the change. A new `config.d/00-native-api-bench.dae` was added, so its routing rules precede the existing process/MAC/geodata rules. WAN is `auto`; LAN interfaces are `wg0,bond0`; `dial_mode: domain`; real eBPF/NFQUEUE. Existing configured proxy groups and geodata were retained. The complete private node/subscription files are intentionally omitted.

New include (entire file; lab node has no credentials):

```dae
node {
    native-api-bench-peer: 'socks5://198.18.80.2:11083'
}
group {
    native-api-benchmark {
        filter: name('native-api-bench-peer')
        policy: fixed(0)
    }
}
routing {
    sip(198.18.81.2/32) && dip(198.18.80.2/32) && dport(15201, 18081, 15203) -> native-api-benchmark(must)
    sip(198.18.81.2/32) && dip(198.18.80.2/32) && dport(15202) -> direct(must)
}
```

Existing routing file, unchanged during the failure:

```dae
# 路由规则
# 参考: https://github.com/daeuniverse/dae/blob/main/docs/routing.md


routing {
    # ============================================================
    # must_direct: DNS 服务、代理客户端、系统服务
    # ============================================================
    pname(dnsmasq, systemd-resolved) && l4proto(udp) && dport(53) -> direct(must)
    pname(mosdns, honk-subsribe, honk-tool) -> direct(must)
    pname(NetworkManager) -> direct(must)
[...]

PR #286, opened 2026-09-21
fix(routing): resolve facts at rule entry instead of behind a readiness flag

Zakkaus:
## Problem

The merged form of #281 resolves the destination, source and MAC facts at their first reached use behind a per-invocation readiness mask in `R8` (`crates/honk-core/src/control/routing_matcher/codegen.rs:466-487` on `origin/main`), with port conditions reordered ahead of the other conditions so a failing port skips the lookups. That mask is a precise constant that differs per path, the `and; jne` on it is a predictable branch, and states with different masks cannot subsume one another, so the verifier revisits every later process-name chain per mask. Measured on Linux 6.12.107 with the same instrumented loader as #281: the #280 policy costs 160,415 processed instructions on `main` against 53,380 with rule-entry resolution, and a policy with four fact rules ahead of twenty-nine `pname` rules 368,874 against 100,191 (`max_states_per_insn` 14 vs 4). The multiplier grows with the number of categories resolved conditionally ahead of the chains, so a somewhat larger policy of the #280 shape exceeds the budget again; `.agents/rules/real-ebpf.md` recorded that the lazy control flow was unmeasured.

## How I fixed it

Facts are resolved once, unconditionally, at the entry of the first rule that uses them (as in the #281 submission): the readiness mask, `emit_fact_bit_lazy`, `or_imm` and the port-first reordering go away, the bitmap copy/zero-fill and every predicate emitter are unchanged, and the resolved set at every rule entry is the same on every path. The cost is at most three extra fact-map lookups per routing evaluation on flows whose rule would have short-circuited on a port, and only from the first rule that uses the category; a routing evaluation runs on flow-cache misses, not on ordinary cached packets. The design docs (en/zh) and the rule file now state the constraint and the measurements. The three `lazy_fact_cache_*` kernel tests are renamed and their stale comments corrected; assertions unchanged.

## Verified

Lab VM (Debian 13, Linux 6.12.107, release `--features ebpf`): the table for both emitters over seven policy variants is at https://paste.gentoozh.org/md/BggThr05m9p (this PR: 53,279 / 53,380 / 100,191 for without / with / stress, `max_states_per_insn` 4 throughout). `just test-routing` on that VM: 32 passed, including `verifier_budget` and `predicate_semantics`. `codegen` unit tests (`--features ebpf`): 7 passed; the three restored rule-entry tests are red on the lazy emitter, the maintainer's copy/zero-fill layout test is kept. `cargo clippy -p honk-core --features ebpf --all-targets -- -D warnings`, `cargo clippy --all --all-targets -- -D warnings`, `cargo fmt --all -- --check` and `cargo test --workspace` clean. One gpt-6 diff review (merge with edits; both edits applied: narrowed verifier wording, renamed/corrected the kernel tests).

---

- [x] `just fmt`, `just lint` and `just test-ci` pass (`test-ci` run as `cargo test --workspace`).


Zakkaus:
Superseded by #287: same measurements, but it keeps the lazy design and only changes what the verifier sees of the readiness mask.

PR #287, opened 2026-09-21
fix(routing): keep the readiness mask opaque to the verifier

Zakkaus:
## Problem

The native routing emitter resolves the destination, source and MAC facts lazily at their first reached use behind a per-invocation readiness mask in `R8` (`crates/honk-core/src/control/routing_matcher/codegen.rs:466-487` on `origin/main`). The mask starts as `mov r8, 0`, so it is a precise constant whose value differs per resolution history, the guard on it is a predictable branch, and states with different values cannot subsume one another: the verifier walks every later process-name chain once per value. Measured on Linux 6.12.107 with the instrumented loader from #281: the #280 policy costs 160,415 processed instructions on `main` (`max_states_per_insn` 9) and a policy with four fact rules ahead of twenty-nine `pname` rules 368,874 (14), where the same policies cost 53k / 100k without the split. The multiplier grows with the number of categories resolved conditionally ahead of the chains, so a somewhat larger policy of the #280 shape exceeds the budget again; `.agents/rules/real-ebpf.md` recorded that the lazy control flow was unmeasured.

## How I fixed it

The lazy design stays; the mask is kept out of the verifier's sight. The prologue loads `R8` back from the decision's just-zeroed `mark` (a memory load the verifier does not fold into the stored constant; zero at runtime) and pre-fills the three lazily resolved areas from fresh loads of the same field, so a use verified before its resolution reads initialized stack and the stored words share no scalar id with the mask. Each guard becomes one `jset` on the mask register itself, which refines the tested bit on both edges so the skip edge and the resolve edge agree at the bit test; the resolution follows in line and publishes the bit. Three instructions fewer per use site, twenty-five more in the prologue. Both design docs and the rule file describe the three properties and what breaks each of them (a constant mask, areas spilled from `R8`, a guard on a copy of the mask), with the measurements.

## Verified

Lab VM (Debian 13, Linux 6.12.107, release `--features ebpf`), every variant measured on the same seven policies: https://paste.gentoozh.org/md/Ehp2upmgrqa. This PR: 53,383 / 53,741 / 100,645 processed instructions for without / with / stress, `max_states_per_insn` 4 throughout; `jset` alone on the constant mask changes nothing (160,097), the opaque mask with the old guard 82,234, areas spilled from `R8` 133,354. `just test-routing` on that VM: 32 passed, including `verifier_budget` and `predicate_semantics`. `codegen` unit tests (`--features ebpf`): 8 passed; `readiness_mask_and_lazy_areas_start_as_verifier_unknown_zeros` pins the prologue shape (mark zeroed, not rewritten, mask and twelve area words loaded from it) and the rewritten guard test pins `jset` on `R8`, the in-line resolution and the publish before the bit test. `cargo clippy -p honk-core --features ebpf --all-targets -- -D warnings`, `cargo clippy --all --all-targets -- -D warnings`, `cargo fmt --all -- --check` and `cargo test --workspace` clean. Two gpt-6 diff review rounds; the second's edits (edge-order wording, narrowed opacity claims, consistent numbers) are applied. Supersedes #286.

---

- [x] `just fmt`, `just lint` and `just test-ci` pass (`test-ci` run as `cargo test --workspace`).


Glassyiris (approved):
ＬＧＴＭ

Zakkaus:
> ＬＧＴＭ

感受到了嗎

